PR #4, opened 2024-04-08
[DPE-2980] credentials rotation

juditnovak:
Verifying that credentials are rotated on the relaton, once changed on `opensearch-dashboards` side.


Mehdi-Bendriss (commented):
Thank you Judit. Should we add a hook for `authentication_updated`?

juditnovak:
We don't need to. IT's redundant. (Secret changed triggers `authentication_updated`)

Mehdi-Bendriss (commented):
This PR LGTM - should be merged as soon as https://github.com/canonical/opensearch-operator/pull/216 is merged on the opensearch side

juditnovak:
@Mehdi-Bendriss 

We have a chicken and egg problem here.

As https://github.com/canonical/opensearch-operator/pull/216 is equally using this code.
The point is that that modification doesn't work without this one and vica versa.
I.e. that PR (just like this one) has a complete copy of this particular version of this charm in it.

So I suggest we rather "mess up" (i.e. add another charm's code to the integration test's space -- not a major "messup" really) for a couple of day's time the (still immature) Dashboard charm, rather then the Opensearch repo.

Mehdi-Bendriss (approved):

PR #19, opened 2024-04-27
[DPE-4033] Cleaning up the TLS workflow

juditnovak:
The TLS workflow (as copied over from another charm) had a few extra solutions not required for the Dasboard.

The workflow now is simplified and adjusted to our needs.

Furthermore additional status changes are introduced. Typically when re-configuring.


tests/integration/test_charm.py
```diff
@@ -176,16 +166,27 @@ async def test_dashboard_password_rotation(ops_test: OpsTest):
...
-        assert access_dashboard_https(host=host, password=new_password)
+    # We need to face it:
+    # 1. updating a credential in a charm (Opensearch)
+    # 2. that has to be updated on a relation (Opensearch <-> Dashboards)
+    # 3. so that another charm could take it over it its configuration
+    # ...takes time...
+    # And 'active' status is not a reliable indicator, given the elaborate
+    # chain of actions to be invoked (where in-between events processing it
+    # seems like there IS a moment where everyone may be idle and active
+    # before/after various handler executions, as part of the chain).
```
juditnovak, line 183: This should be addressed here https://warthogs.atlassian.net/browse/DPE-4226 however I'm suspicious that the `Retry` below still needs to be kept. Simply because there may be a delay until the machinery eventually gets through the process, and there may be still potential gaps in time when "accidentally" all apps/units may be active for a moment.

src/charm.py
```diff
@@ -129,29 +131,46 @@ def reconcile(self, event: EventBase) -> None:
...
 
-        if event.secret.label == self.state.cluster.data_interface._generate_secret_label(
+        cluster_secret_label = self.state.cluster.data_interface._generate_secret_label(
             PEER,
-            self.state.cluster.relation.id,
-            'extra',  # type:ignore noqa
-        ):  # Changes with the soon upcoming new version of DP-libs STILL within this POC
+            self.state.peer_relation.id,
+            "extra",  # type:ignore noqa
+        )  # Changes with the soon upcoming new version of DP-libs STILL within this POC
```
Mehdi-Bendriss, line 165: Comments still relevant?

src/literals.py
```diff
@@ -44,3 +44,5 @@
 MSG_WAITING_FOR_USER_CREDENTIALS = "waiting for passwords to be created"
 MSG_WAITING_FOR_PEER = "waiting for peer relation"
 MSG_DB_MISSING = "Opensearch connection is missing"
+MSG_TLS_CONFIG = "Waiting for TLS is to be fully configured..."
```
Mehdi-Bendriss, line 47:
```suggestion
MSG_TLS_CONFIG = "Waiting for TLS to be fully configured..."
```

tests/integration/helpers.py
```diff
@@ -161,6 +161,28 @@ def access_dashboard_https(host: str, password: str):
...
+    dashboard_credentials = await get_secret_by_label(
+        ops_test, f"opensearch-client.{relation_id}.user.secret"
+    )
+    dashboard_password = dashboard_credentials["password"]
+
+    # Copying the Dashboard's CA cert locally to use it for SSL verification
+    # We only get it once for pipeline efficiency, as it's the same on all units
+    if https:
+        unit = ops_test.model.applications[APP_NAME].units[0].name
+        get_dashboard_ca_cert(ops_test.model.name, unit)
```
Mehdi-Bendriss, line 176:
```suggestion
        assert get_dashboard_ca_cert(ops_test.model.name, unit), "CA certificates missing."
```

tests/integration/test_charm.py
Mehdi-Bendriss: I believe you can have this done using an acceptably high enough `idle_period=30` in the above `wait_for_idle`.

Mehdi-Bendriss (commented):
Thanks Judit - I have a couple of comments

src/charm.py
```diff
@@ -129,29 +131,46 @@ def reconcile(self, event: EventBase) -> None:
         if getattr(event, "departing_unit", None) == self.unit:
             return
 
+        # Maintain the correct status
+        if self.unit.is_leader():
+            if self.state.opensearch_server:
+                clear_status(self.app, MSG_DB_MISSING)
+
+            if self.state.cluster.tls and not self.state.cluster.tls:
```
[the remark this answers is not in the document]
juditnovak: 2nd one is supposed to be `self.state.unit_server.tls`, fixing.

src/charm.py
```diff
@@ -129,29 +131,46 @@ def reconcile(self, event: EventBase) -> None:
...
+                self.unit.status = MaintenanceStatus(MSG_TLS_CONFIG)
```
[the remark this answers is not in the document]
juditnovak: Let me evalue this (I'm not yet convenienced). More news soon.

src/charm.py
```diff
@@ -129,29 +131,46 @@ def reconcile(self, event: EventBase) -> None:
...
+            else:
+                clear_status(self.unit, MSG_TLS_CONFIG)
+
+        self.unit.status = MaintenanceStatus(MSG_RECONFIGURING)
```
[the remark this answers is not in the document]
juditnovak: Re-configuration occurs within the condition.

Now, even if we do not end up re-configuring (as config is correct) we still make an attempt.
Is there an issue with signaling that we're checking our config and perhaps adjust it if needed?
This was the intent of this message.

In other words: it's deep down in checking for config changes, when this decision is taken.
I don't think it's good to spread around status changes across the code. 
I would like to keep them in a consize, well-maintainable space.

Which is the reason why I choose to signal on "Re-configuration attemtp", and not only "Reconfiguration" strict.

src/charm.py
juditnovak: Yes.

tests/integration/helpers.py
juditnovak: I would like to avoid helpers within `helpers` raising `assert` failures.

The tests will fail anyway if CA cert can't be retieved.

tests/integration/test_charm.py
juditnovak: Let me give it a try.

tests/integration/helpers.py
juditnovak: Bah, fine, taking it onboard, it increases verbosity

src/charm.py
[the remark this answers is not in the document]
juditnovak: ...meaning: TLS relation is established, but the unit didn't fetch its certificates yet.

tests/integration/test_charm.py
juditnovak: Seems working.

I leave it as such for now. I'm running these pipelines a lot, if it seems unstable I'll bring back the retry

src/charm.py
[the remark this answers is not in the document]
Mehdi-Bendriss: On the **leader** unit, the flow would be:
1. unit.status = MaintenanceStatus(MSG_TLS_CONFIG)
2. unit.status = MaintenanceStatus(MSG_RECONFIGURING)

In other words, "1." will never be displayed. Did you perhaps mean to set the status to the application instead of the unit?

phvalguima (approved):

src/charm.py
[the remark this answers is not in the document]
juditnovak: You are right.

Since each reconfiguration actually requires a restart, perhaps it's pointless to keep that status change.
Restrarts anyway set corresponding statuses for each step.

Thus I guess it's best to drop the `MSG_RECONFIGURING` status.

src/charm.py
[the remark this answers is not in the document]
juditnovak: IT's fine, let's just drop it. It's enough to notify about restarts. (Re)-confguration attempts after all aren't particularly important for the users to know.

src/charm.py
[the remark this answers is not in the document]
juditnovak: Regarding `return`: I've opened this issue https://github.com/canonical/opensearch-dashboards-operator/issues/21 to signify that currently (given the low number of occurances combined with the fast restart speed -- all of which should anyway be hidden behind `traefik`) we choose to restart on the few occasions needed. However may the number grow, this design is to be reviewed.

Mehdi-Bendriss (approved):

phvalguima (approved):

PR #20, opened 2024-04-28
[DPE-2947] traefik connection

juditnovak:


juditnovak:
Traefik doesn't support VM connection

PR #23, opened 2024-05-01
[DON'T MERGE][DPE-2983] HA network cut stress test

juditnovak:


juditnovak:
Skipped, not needed (HA tetss are very stable, "never beraks")

PR #25, opened 2024-05-04
[JUJU SECRET DEBUG][DPE-2983] ha network cut

juditnovak:
Pipeline to remonstrate missing (or missed?) `secrt-changed` event for Juju bug


juditnovak:
This POC is not needed anymore, as the issue have been tracked down and addressed in https://github.com/canonical/opensearch-dashboards-operator/pull/51

PR #22, opened 2024-04-29
[DPE-2983] HA network cut

juditnovak:
NOTE to Reviewers:

Plese take a look at "Usage Instructions" in [this comment](https://github.com/canonical/opensearch-dashboards-operator/pull/22#issuecomment-2092721246) before reviewing this PR

Shotrly:
When reviewing changs, in the Files (left-hand side) tab
 - close the `libs` tab (nothig to review there)
 - close the `tests/integration/opensearch-operator` tab (nothing to review here)
 
 Furthemore, you may find review by commit handy.

PR #26, opened 2024-05-06
[DPE-3954] Final cleanup

juditnovak:
Response to the open questions in the related ticket [DPE-3954](https://warthogs.atlassian.net/browse/DPE-3954):

COS integration:
 - Kibana has an (outdated (*)) [Prometheus integrator module]( https://github.com/pjhampton/kibana-prometheus-exporter), which --exactly as for Opensearch-- requires specific user that can connect to the Dashboard. From what I undrestand from the context, this is rather to be a user of the Dasboard, and not an Opensearch user. 
   - Overall it seems to me as a potential source of duplicating work in the near future, if we remove local users from the Dashboard, we should rather keep it
   - (*) NOTE: If the outdated exporter module is replaced with a new one, it's more than likely to use a very similar way of access
 
- adopting Openserach YAML setter (see https://github.com/canonical/opensearch-dashboards-operator/pull/1#discussion_r1537252675))
  - Still no particular need for that, can be done if a functionality requires this specific implementation. Until that moment there is no harm in holding on to the current 'keep-it-simple' flat implementation

[DPE-3954]: https://warthogs.atlassian.net/browse/DPE-3954?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


phvalguima (commented):
@juditnovak I recall we discussed [also removing this code](https://github.com/canonical/opensearch-dashboards-operator/blob/ef6cbbcdede47b4e49b668a72d7299ea0fea0684/tests/integration/helpers.py#L186) as it is left over from ZK charm

juditnovak:
@phvalguima Good call, removing.

juditnovak:
@phvalguima All right, I knew something was wrong here ;-)

Are you sure you were checking for the fucntion on _this_ branch?

IT was removed in this base branch ;-)

https://github.com/canonical/opensearch-dashboards-operator/pull/22/files#diff-e9349e06331f9876f984bcf05671235cfbe546046bccab6ad1abb0b183982b30L186-L284

phvalguima (approved):
LGTM

PR #55, opened 2024-05-29
[DPE-4397] Rolling upgrades (v1)

juditnovak:
Rolling upgrades (v1) for Opensearch Dashboards.

### DIFF
--- a/src/events/upgrade.py
+++ b/src/events/upgrade.py
@@ -1,0 +1,100 @@
+#!/usr/bin/env python3
+# Copyright 2023 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""Event handler for handling OpensearchDashboards in-place upgrades."""
+import logging
+import time
+from typing import TYPE_CHECKING
+
+from charms.data_platform_libs.v0.upgrade import (
+    BaseModel,  # pyright: ignore [reportGeneralTypeIssues, reportAttributeAccessIssue]
+)
+from charms.data_platform_libs.v0.upgrade import (
+    ClusterNotReadyError,
+    DataUpgrade,
+    DependencyModel,
+    UpgradeGrantedEvent,
+)
+from typing_extensions import override
+
+if TYPE_CHECKING:
+    from charm import (
+        OpensearchDashboardsCharm,  # pyright: ignore [reportGeneralTypeIssues, reportAttributeAccessIssue]
+    )
+
+logger = logging.getLogger(__name__)
+
+
+class OpensearchDashboardsDependencyModel(BaseModel):
+    """Model for OpensearchDashboards Operator dependencies."""
+
+    osd_upstream: DependencyModel
+
+
+class ODUpgradeEvents(DataUpgrade):
+    """Implementation of :class:`DataUpgrade` overrides for in-place upgrades."""
+
+    def __init__(self, charm: "OpensearchDashboardsCharm", **kwargs):
+        super().__init__(charm, **kwargs)
+        self.charm = charm
+
+    def post_upgrade_check(self) -> None:
+        """Runs necessary checks validating the unit is in a healthy state after upgrade."""
+        self.pre_upgrade_check()
+
+    @override
+    def pre_upgrade_check(self) -> None:
+        pass
+
+    @override
+    def build_upgrade_stack(self) -> list[int]:
+        upgrade_stack = []
+        units = [self.charm.unit] + list(self.charm.state.peer_relation.units)
+        for unit in units:
+            upgrade_stack.append(int(unit.name.split("/")[-1]))
+
+        return upgrade_stack
+
+    @override
+    def log_rollback_instructions(self) -> None:
+        logger.critical(
+            "\n".join(
+                [
+                    "Unit failed to upgrade and requires manual rollback to previous stable version.",
+                    "    1. Re-run `pre-upgrade-check` action on the leader unit to enter 'recovery' state",
+                    "    2. Run `juju refresh` to the previously deployed charm revision",
+                ]
+            )
+        )
+        return
+
+    @override
+    def _on_upgrade_granted(self, event: UpgradeGrantedEvent) -> None:
+        self.charm.workload.stop()
+
+        if not self.charm.workload.install():
+            logger.error("Unable to install OpensearchDashboards...")
+            self.set_unit_failed()
+            return
+
+        logger.info(f"{self.charm.unit.name} upgrading workload...")
+        self.charm.workload.restart()
+
+        time.sleep(5.0)
+
+        try:
+            logger.debug("Running post-upgrade check...")
+            self.post_upgrade_check()
+
+            logger.debug("Marking unit completed...")
+            self.set_unit_completed()
+
+            # ensures leader gets it's own relation-changed when it upgrades
+            if self.charm.unit.is_leader():
+                logger.debug("Re-emitting upgrade-changed on leader...")
+                self.on_upgrade_changed(event)
+
+        except ClusterNotReadyError as e:
+            logger.error(e.cause)
+            self.set_unit_failed()
--- a/src/events/upgrade.py
+++ b/src/events/upgrade.py
@@ -4,7 +4,6 @@
 
 """Event handler for handling OpensearchDashboards in-place upgrades."""
 import logging
-import time
 from typing import TYPE_CHECKING
 
 from charms.data_platform_libs.v0.upgrade import (
@@ -16,7 +15,10 @@ from charms.data_platform_libs.v0.upgrade import (
     DependencyModel,
     UpgradeGrantedEvent,
 )
+from ops.model import BlockedStatus
 from typing_extensions import override
+
+from literals import MSG_INCOMPATIBLE_UPGRADE
 
 if TYPE_CHECKING:
     from charm import (
@@ -41,11 +43,24 @@ class ODUpgradeEvents(DataUpgrade):
 
     def post_upgrade_check(self) -> None:
         """Runs necessary checks validating the unit is in a healthy state after upgrade."""
-        self.pre_upgrade_check()
+        srv_version_actual = self.charm.state.opensearch_server.version
+        srv_version_required = self.dependency_model.osd_upstream.dependencies["opensearch"]
+        major_actual, minor_actual = srv_version_actual.split(".")[:2]
+        major_required, minor_required = srv_version_required.split(".")[:2]
+        if major_actual > major_required or minor_actual > minor_required:
+            self.charm.unit.status = BlockedStatus(MSG_INCOMPATIBLE_UPGRADE)
+            raise ClusterNotReadyError(
+                message="Post-upgrade check failed and cannot safely upgrade",
+                cause="Opensearch version mismatch",
+            )
 
     @override
     def pre_upgrade_check(self) -> None:
-        pass
+        if not self.charm.workload.alive():
+            raise ClusterNotReadyError(
+                message="Pre-upgrade check failed and cannot safely upgrade",
+                cause="Unit workload is not running",
+            )
 
     @override
     def build_upgrade_stack(self) -> list[int]:
@@ -75,13 +90,11 @@ class ODUpgradeEvents(DataUpgrade):
 
         if not self.charm.workload.install():
             logger.error("Unable to install OpensearchDashboards...")
-            self.set_unit_failed()
+            self.set_unit_failed(cause="Workload install failed")
             return
 
         logger.info(f"{self.charm.unit.name} upgrading workload...")
         self.charm.workload.restart()
-
-        time.sleep(5.0)
 
         try:
             logger.debug("Running post-upgrade check...")
@@ -97,4 +110,4 @@ class ODUpgradeEvents(DataUpgrade):
 
         except ClusterNotReadyError as e:
             logger.error(e.cause)
-            self.set_unit_failed()
+            self.set_unit_failed(cause=e.cause)
--- a/src/literals.py
+++ b/src/literals.py
@@ -19,6 +19,15 @@ CONTAINER = "opensearch-dashboards"
 CHARM_USERS = ["monitor"]
 SERVER_PORT = 5601
 
+DEPENDENCIES = {
+    "osd_upstream": {
+        "dependencies": {},
+        "name": "opensearch-dashboards",
+        "upgrade_supported": ">=2",
+        "version": "2.12",
+    },
+}
+
 PATHS = {
     "CONF": "/var/snap/opensearch-dashboards/current/etc/opensearch-dashboards",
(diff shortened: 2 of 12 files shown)


src/events/upgrade.py
```diff
@@ -0,0 +1,100 @@
...
+        super().__init__(charm, **kwargs)
+        self.charm = charm
+
+    def post_upgrade_check(self) -> None:
+        """Runs necessary checks validating the unit is in a healthy state after upgrade."""
+        self.pre_upgrade_check()
+
+    @override
+    def pre_upgrade_check(self) -> None:
+        pass
```
juditnovak, line 48: I'm wondering if we want to have a pre-post upgrade check.

In fact, what I would suggest is to leave this empty at the moment, and get back to this in https://warthogs.atlassian.net/browse/DPE-4226, where we are planning to implement a service healthcheck. I think that's the thing that should be executed before and after an upgrade.

tests/integration/test_upgrade.py
```diff
@@ -0,0 +1,143 @@
...
+    assert "upgrade-stack" in get_relation_data(
+        model_full_name=ops_test.model_full_name, unit=f"{APP_NAME}/0", endpoint="upgrade"
+    )
+
+    await ops_test.model.applications[APP_NAME].refresh(path=pytest.charm)
+    await ops_test.model.wait_for_idle(
+        apps=[APP_NAME], status="active", timeout=1000, idle_period=120
+    )
+
+    assert await access_all_dashboards(ops_test)
```
juditnovak, line 104: I've tested manually, the upgrade process seems fully functional.
I'm saying this because the tests -- right now -- can not confirm that :-)
(They only show that after running a update we have a functional service :-) )

What I mean is that in my next PR (i.e. after we already have a release out using the `upgrades` lib) I will add here an actual version check, to verify that an upgrade actually happened.

src/events/upgrade.py
```diff
@@ -0,0 +1,100 @@
...
+
+    @override
+    def build_upgrade_stack(self) -> list[int]:
```
phvalguima, line 51: if we are following the process as OpenSearch, this is not needed anymore.

phvalguima (commented):
Maybe I don't have 100% of the context here. However, my recommendation is to follow a process that is closer to OpenSearch's: https://github.com/canonical/opensearch-operator/blob/main/src/upgrade.py

src/events/upgrade.py
juditnovak: As for now, we are not following the Opensearch solution It has two reasons.

Within this code I have "inherited" a full solution for v1 -- which makes this change really a low-hanging fruit ;-)
Moving from a functional 'upgades v1' is to 'v2' is easier than (re-)implementing all from scratch. Not to mention that v1 already provides a fully functional upgrades feature to the Dashboards.

In the meanwhile, we may not even need v2 for the Dashboards, as --after all-- we are dealing with a lightweight, stateless charm here. 

Alltogether: it's either a simple temporary change delivering quite some value, or may even be the permanent solution.

src/events/upgrade.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+class ODUpgradeEvents(DataUpgrade):
+    """Implementation of :class:`DataUpgrade` overrides for in-place upgrades."""
+
+    def __init__(self, charm: "OpensearchDashboardsCharm", **kwargs):
+        super().__init__(charm, **kwargs)
+        self.charm = charm
+
+    def post_upgrade_check(self) -> None:
```
phvalguima, line 42: I recommend we add a `workload_version` file in this charm. At `post_upgrade_check`, you can check the version of your snap currently installed vs. your workload. If this is not, e.g. a major upgrade, then we can proceed.

phvalguima (commented):
Left some comments. One general question I have is where `charm_version` file is actually used. If we do not, then I think we can remove it.

src/events/upgrade.py
juditnovak: Regarding Opensearch upgrades:
We are putting a requirement on the Dashboards Application version (which is linked to an Opensearch version).
However that's indeed not a dynamic cross-service check.

Question #1: Do we want to add this feature (here)? I recall we both had a taks on aligning Opensearch and Dashboards versions (but I'm struggling to find it in Jira now). At the time we decided not to get that task done.

Question #2: According to [Charm Interfaces](https://github.com/canonical/charm-relation-interfaces/tree/main/interfaces/opensearch_client/v0), currently Opensearch is not publishing vesion information. At a quick check I couldn't find a corresponding API call either (but I may be missing something at that level). How could we check the version or update status of Opensearch?

Regarding "green" state:
Yes, I agree. However the "green" services healthcheck is to be implemented within the scope of the task mentioned in my comment. So I think to re-visit the `pre_upgrade_check` after the dependency task is completed.

src/literals.py
```diff
@@ -19,6 +19,15 @@
 CHARM_USERS = ["monitor"]
 SERVER_PORT = 5601
 
+DEPENDENCIES = {
+    "osd_upstream": {
```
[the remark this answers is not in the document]
juditnovak: Good point, adding it!

src/events/upgrade.py
```diff
@@ -0,0 +1,100 @@
...
+        )
+        return
+
+    @override
+    def _on_upgrade_granted(self, event: UpgradeGrantedEvent) -> None:
+        self.charm.workload.stop()
+
+        if not self.charm.workload.install():
+            logger.error("Unable to install OpensearchDashboards...")
+            self.set_unit_failed()
```
Mehdi-Bendriss, line 78: This seems to set an empty `BlockedStatus` on the unit, we should probably pass a `cause`

src/events/upgrade.py
```diff
@@ -0,0 +1,100 @@
...
+            return
+
+        logger.info(f"{self.charm.unit.name} upgrading workload...")
+        self.charm.workload.restart()
+
+        time.sleep(5.0)
```
Mehdi-Bendriss, line 84: This should move to the `workload.restart` logic, and we should be less conservative about the number of seconds, we can have a retry mechanism every second until it's up - which should ideally take less than 5s

src/events/upgrade.py
```diff
@@ -0,0 +1,100 @@
...
+
+            logger.debug("Marking unit completed...")
+            self.set_unit_completed()
+
+            # ensures leader gets it's own relation-changed when it upgrades
+            if self.charm.unit.is_leader():
+                logger.debug("Re-emitting upgrade-changed on leader...")
+                self.on_upgrade_changed(event)
+
+        except ClusterNotReadyError as e:
```
Mehdi-Bendriss, line 98: When would this exception be raised? I assume on `post_upgrade_check`?

src/events/upgrade.py
```diff
@@ -0,0 +1,100 @@
...
+            logger.error(e.cause)
+            self.set_unit_failed()
```
Mehdi-Bendriss, line 100: We should probably pass the cause to this method

src/literals.py
```diff
@@ -19,6 +19,15 @@
...
+        "dependencies": {},
+        "name": "opensearch-dashboards",
+        "upgrade_supported": ">=2",
+        "version": "2.12",
```
Mehdi-Bendriss, line 27: FYI - we are at 2.14

tests/integration/test_upgrade.py
```diff
@@ -0,0 +1,143 @@
...
+    for unit in ops_test.model.applications[APP_NAME].units:
+        if await unit.is_leader_from_status():
+            leader_unit = unit
+    assert leader_unit
+
+    action = await leader_unit.run_action("pre-upgrade-check")
+    await action.wait()
+
+    # ensure action completes
+    time.sleep(10)
```
Mehdi-Bendriss, line 92: this should not be needed, `action.wait` is a blocking operation

tests/integration/test_upgrade.py
```diff
@@ -0,0 +1,143 @@
...
+
+    # ensuring app is safe to upgrade
+    assert "upgrade-stack" in get_relation_data(
```
Mehdi-Bendriss, line 95: Can we add a check whether the upgrade stack is filled with the expected unit ids?

tests/integration/test_upgrade.py
```diff
@@ -0,0 +1,143 @@
...
+    for unit in ops_test.model.applications[APP_NAME].units:
+        if await unit.is_leader_from_status():
+            leader_unit = unit
+    assert leader_unit
+
+    action = await leader_unit.run_action("pre-upgrade-check")
+    await action.wait()
+
+    # ensure action completes
+    time.sleep(10)
```
Mehdi-Bendriss, line 131: same as above - not needed

tests/integration/test_upgrade.py
```diff
@@ -0,0 +1,143 @@
...
+
+    # ensuring app is safe to upgrade
+    assert "upgrade-stack" in get_relation_data(
```
Mehdi-Bendriss, line 134: same as above - can we add a check on the content of the stack?

Mehdi-Bendriss (commented):
Thanks Judit, I left some comments and questions

src/events/upgrade.py
juditnovak: I see your point.

Should we have infinite or finite amount of retries? I mean we keep tying until the node is up, or we keep trying N times?

juditnovak: ("until the node is up" sounds a bit dangerous to me :sweat_smile: )

src/events/upgrade.py
juditnovak: TBH in our case it's just to comply to the "usual template".

Normally `ClusterNotReadyError` is the one that's recommended to raise during the `pre_upgrade_check`.

So in case we add a `pre_upgrade_check` instead of the current empty one, that's when this could make sense.

src/events/upgrade.py
juditnovak: Summarizing live discussions: we said we would block on upgrading to a higher workload version than what Opensearch may have.

Yes, I also found it in the meanwhile that we communicate the vesion in the relation, and that's what I'm planning to use for the check. 

However a related question (outside of this PR): shouldn't we update [Charm Interfaces](https://github.com/canonical/charm-relation-interfaces/tree/main/interfaces/opensearch_client/v0)?

src/literals.py
juditnovak: Yes, I noticed the Opensearch upgrade.

I'm adding tickets to upgrade the Dashboards as well when I'm back.

juditnovak: - [DPE-4565](https://warthogs.atlassian.net/browse/DPE-4565) [OpenSearch Dashboards][VM] - Upgrade - Snap v2.13
- [DPE-4566](https://warthogs.atlassian.net/browse/DPE-4566) [OpenSearch Dashboards][VM] - Upgrade - Snap v2.14
- [DPE-4567](https://warthogs.atlassian.net/browse/DPE-4567) [OpenSearch Dashboards][VM] - Upgrade - Upgrade - Charm workload version v2.13 
- [DPE-4568](https://warthogs.atlassian.net/browse/DPE-4568) [OpenSearch Dashboards][VM] - Upgrade - Upgrade - Charm workload version v2.14

src/events/upgrade.py
Mehdi-Bendriss: until the node is up, but before the set timeout which can be something like 30sec for ex

src/literals.py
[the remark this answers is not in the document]
juditnovak: Done.

src/events/upgrade.py
juditnovak: To summarize a live discussion, we found that this "upgrades v2" feature is rather to be used when moving to "v2"

src/events/upgrade.py
juditnovak: I've added a tenacity retry around the `workload.restart()` function.
After each restart attempt it's checking using `workload.alive()` (which also has tenacity retry around it).

Do you think this may be sufficient?

src/events/upgrade.py
juditnovak: Update: Having added a `pre_upgrade_check` and a `post_upgrade_check`, this catch is not only for "following protocols", but to take action on corresponding failure cases.

src/events/upgrade.py
juditnovak: Done.

tests/integration/test_upgrade.py
juditnovak: Done.

tests/integration/test_upgrade.py
juditnovak: Done.

tests/integration/test_upgrade.py
juditnovak: Done.

tests/integration/test_upgrade.py
juditnovak: Done.

src/events/upgrade.py
phvalguima: nit: `post_upgrade_check` should throw a different error, as the "cluster is ready".

phvalguima (approved):
Left one last comment but I think we can address in separated if we are not changing this PR anymore.

src/events/upgrade.py
phvalguima: @juditnovak approved the PR but I suggest we keep this comment open as we will likely revisit these discussions at `v2`

juditnovak: Sure. Thank you.

juditnovak: Just to note on this thread: the current mechanism surely works :rofl: 

https://github.com/canonical/opensearch-dashboards-operator/actions/runs/9414634059/job/25939480084

Pinning down Opensearch in the pipelines to the last 2.12 releasse :-)

Mehdi-Bendriss: > shouldn't we update [Charm Interfaces](https://github.com/canonical/charm-relation-interfaces/tree/main/interfaces/opensearch_client/v0)?

Absolutely! We should add the `kibanaserver` role there

Mehdi-Bendriss (approved):

phvalguima (approved):

src/events/upgrade.py
juditnovak: ...together with the vresion number being exposed.

I craete a ticket so we dont' foget.

src/events/upgrade.py
juditnovak: OK, I could turn that one to a better one. I thnk I saw a version mismetch exception in the `upgades` lib.

Ill do the change within another PR (ast itls really small).

src/events/upgrade.py
juditnovak: https://warthogs.atlassian.net/browse/DPE-4576

PR #81, opened 2024-07-03
[DPE-4568] Version upgrade to v2.14

juditnovak:
Opensearch Dashboards v2.14 (snap revision 10) adopted to the charm code.

(Given that charmlibs got oudtated, we also do an update for those.)


phvalguima (approved):

Mehdi-Bendriss (approved):
Could you add a title as to what is this PR fixing as well as a short description?

PR #82, opened 2024-07-03
[DPE-4767] HA network cut tests stabilization

juditnovak:
When switching on TLS (halfway through the HA network cut tests) sometimes we try to access the service premature, i.e. before it's fully functional. 

This PR is an attempt to fix that.

NOTE: This PR went through a few stages since originally opened (more details on the related Jira ticket).

A change in behavior got implemented here, namely removing support for switching back-and-forth between HTTP and HTTPS. From now on, if the charm was **once** related to the `self-signed-certificates` operator, it is NOT possible to switch it back to use HTTP protocol ever again.

### DIFF
--- a/src/events/tls.py
+++ b/src/events/tls.py
@@ -52,24 +52,45 @@ class TLSEvents(Object):
 
         self.framework.observe(getattr(self.charm.on, "config_changed"), self._on_config_changed)
 
-    def _on_certs_relation_joined(self, event: RelationJoinedEvent) -> None:
-        """Handler for `certificates_relation_joined` event."""
-        # generate unit private key if not already created by action
+    def _request_certificates(self):
+        """Request brand-new certificates."""
         if not self.charm.state.unit_server.private_key:
             self.charm.state.unit_server.update(
                 {"private-key": generate_private_key().decode("utf-8")}
             )
 
+        logger.debug(
+            "Requesting certificate for: "
+            f"host {self.charm.state.unit_server.host},"
+            f"with IP {self.charm.state.unit_server.sans.get('sans_ip', [])},"
+            f"DNS {self.charm.state.unit_server.sans.get('sans_dns', [])}"
+        )
+
         csr = generate_csr(
             private_key=self.charm.state.unit_server.private_key.encode("utf-8"),
-            subject=self.charm.state.unit_server.host,
+            subject=self.charm.state.unit_server.private_ip,
             sans_ip=self.charm.state.unit_server.sans.get("sans_ip", []),
             sans_dns=self.charm.state.unit_server.sans.get("sans_dns", []),
         )
 
         self.charm.state.unit_server.update({"csr": csr.decode("utf-8").strip()})
-
         self.certificates.request_certificate_creation(certificate_signing_request=csr)
+
+    def _remove_certificates(self):
+        """Cleanup any existing certificates."""
+        if self.charm.state.cluster.tls:
+            self.certificates.request_certificate_revocation(
+                self.charm.state.unit_server.csr.encode("utf-8")
+            )
+        self.charm.state.unit_server.update({"csr": "", "certificate": "", "ca-cert": ""})
+
+        # remove all existing keystores from the unit so we don't preserve certs
+        self.charm.tls_manager.remove_cert_files()
+
+    def _on_certs_relation_joined(self, event: RelationJoinedEvent) -> None:
+        """Handler for `certificates_relation_joined` event."""
+        # generate unit private key if not already created by action
+        self._request_certificates()
 
     def _on_certificate_available(self, event: CertificateAvailableEvent) -> None:
         """Handler for `certificates_available` event after provider updates signed certs."""
@@ -109,14 +130,12 @@ class TLSEvents(Object):
     def _on_config_changed(self, event: EventBase):
         """If system configuration (such as IP) changes, certs have to be re-issued."""
         if self.charm.state.unit_server.tls and not self.charm.tls_manager.certificate_valid():
-            self._on_certificate_expiring(event)
+            self._remove_certificates()
+            self._request_certificates()
 
     def _on_certs_relation_broken(self, _) -> None:
         """Handler for `certificates_relation_broken` event."""
-        self.charm.state.unit_server.update({"csr": "", "certificate": "", "ca-cert": ""})
-
-        # remove all existing keystores from the unit so we don't preserve certs
-        self.charm.tls_manager.remove_cert_files()
+        self._remove_certificates()
 
     def _set_tls_private_key(self, event: ActionEvent) -> None:
         """Handler for `set-tls-privat-key` event when user manually specifies private-keys for a unit."""
--- a/src/events/tls.py
+++ b/src/events/tls.py
@@ -59,6 +59,10 @@ class TLSEvents(Object):
                 {"private-key": generate_private_key().decode("utf-8")}
             )
 
+        if self.charm.state.unit_server.tls and self.charm.tls_manager.certificate_valid():
+            return
+            # OR SHOULD WE REMOVE (incl. REVOKE) AND DELETE OLD FILES (i.e. run self._remove_certificates())
+
         logger.debug(
             "Requesting certificate for: "
             f"host {self.charm.state.unit_server.host},"
@@ -135,7 +139,9 @@ class TLSEvents(Object):
 
     def _on_certs_relation_broken(self, _) -> None:
         """Handler for `certificates_relation_broken` event."""
-        self._remove_certificates()
+        # In case we have valid certificats, we keep them for smooth service function
+        if self.charm.state.unit_server.tls and not self.charm.tls_manager.certificate_valid():
+            self._remove_certificates()
 
     def _set_tls_private_key(self, event: ActionEvent) -> None:
         """Handler for `set-tls-privat-key` event when user manually specifies private-keys for a unit."""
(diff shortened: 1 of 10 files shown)


src/events/tls.py
```diff
@@ -109,14 +130,12 @@ def _on_certificate_expiring(self, _: EventBase) -> None:
...
+            self._remove_certificates()
+            self._request_certificates()
 
     def _on_certs_relation_broken(self, _) -> None:
         """Handler for `certificates_relation_broken` event."""
-        self.charm.state.unit_server.update({"csr": "", "certificate": "", "ca-cert": ""})
-
-        # remove all existing keystores from the unit so we don't preserve certs
-        self.charm.tls_manager.remove_cert_files()
+        self._remove_certificates()
```
reneradoi, line 138: Maybe it would be worth keeping the certs even though the relation might be broken? Otherwise you would immediately not be able to serve requests anymore, even though the certs themselves are still functional and valid.

tests/integration/ha/test_network_cut.py
```diff
@@ -11,10 +11,17 @@
...
+    access_all_dashboards,
+    all_dashboards_unavailable,
+    get_address,
+    get_leader_name,
+)
 
 logger = logging.getLogger(__name__)
 
+logging.basicConfig(format="%(asctime)s %(levelname)-8s %(message)s", datefmt="%Y-%m-%d %H:%M:%S")
+
```
reneradoi, line 24: What's the reason for this? I assume this should be handled by the test runner (in this case, tox), no?

src/events/tls.py
juditnovak: The service is not functional with the old cert, as the IP of the cert doesn't match anymore with the IP if the host.
The followning error:
```
SSLError(SSLCertVerificationError(1, "[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed
IP address mismatch, certificate is not valid for '10.67.147.132'. (_ssl.c:1007)")))
```
is supposed to be raised, if the certificate verification for the connection is correctly performed.

Thus there is no point keeping an invalid cert around.

You can also look at the whole process this way.
This is the moment when `config-changed` event was received to notify the charm about the new IP.
At this point the service is not yet available. A service restoration process has to take place, including:
 - re-configuring the endpoint URL in the charm config
 - in case HTTPS: requesting and receiving new certificates
Until this chain is not fully complete, the service is not available. We shouldn't even try to pretend that it is. It would just cause more trouble :-)

The restoration process is clearly indicated in the charm status, thus noone should expect the charm to be functional until the above is all complete.

tests/integration/ha/test_network_cut.py
juditnovak: Ah, it was a naive attempt to get timestamps on the CI messages. Didn't work, removing.

src/events/tls.py
reneradoi: You're right for the `config-changed` event after a new IP address was assigned to the unit. I don't want to question that. But I was specifically referring to the `_on_certs_relation_broken` handler, where - at least from my point of view - it would make sense to not remove the certs as they should still be valid.

juditnovak: OK, I do see your point.

Now this is conceptual question.

First of all: this part of the workflow is a bit outside of the scope of the PR in the sense that we didn't introduce a change. (I've only re-grouped the code itself, but the `certificates-relation-broken` workflow was left intact.)

Basically the current policy is this.

When a relation is established with the `self-signed-certificates` operator, the OSD charm requests certificates.
If the certs relation is broken, we invalidate TLS credentials.

If we decide to move away from this, that would require modifications on other parts of the workflow. 

(I'm not saying we shouldn't do it -- but I'm sayng that 
 1. it's a major modfication
     - major change on the TLS workflow
     - introduces a higher level of complexity than the original, simplistic approach
     - requiring further tests changes, etc.
 2. probably should be taken out of the scope of this change that has been going on for about 6 weeks already.)

I'm not against going this way, but would encourage to address it separately.

reneradoi: I'm fine with this, I just wanted to bring it up as Opensearch itself stays functional in that case, so maybe the two should be aligned. But it's absolutely fine when you just put it into an issue or "enhancement request" for later decision.

reneradoi (dismissed):

tests/integration/ha/test_network_cut.py
Mehdi-Bendriss: Could we remove?

Mehdi-Bendriss (commented):
Thanks Judit - 2 minor comments

src/events/tls.py
juditnovak: Please note that we need to do more than the 2 lines above.

This requires proper testing. Integration tests included. As we basically want to verify that valid certs are there on a broken relation -- and invalid certs are removed on a broken relation.

For which reason I would like to address the issue outside of a PR extending since 3rd of July (more than 7 weeks as of today).

Partially since such extending tasks don't allow to show progress. Partially because I prefer to work with consize, self-containing tasks.

juditnovak: (Also previous estimations of tickets become inaacuate whenever additional scope is added to the task. This change was not part of the original plan last time we've estimated this ticket. We already agreed on what changes we were planning to do on the Dashboards TLS workflow.)

Mehdi-Bendriss: @juditnovak let's keep it simple please. There is a low hanging fruit for more robustness which is to add this check. It is a 1 liner. No need for integration testing.

juditnovak: So please let's set it clear: what should be the new workflow.

For the definition below we start from a place where a TLS connection has established.

1. `certificates-relation-broken`
     - We keep certificates if any. (We used to delete them.)
2. `certificates-relation-joined`
     a) We do NOT request certificates if there are any on the unit. (We used to request certificates at this point.)
     b) We DO request certificates if none was found on the unit.

Is this what we would like to have?
(If yes, then I honestly believe that it is a significant change in the workflow -- and there should be an integration test confirming that the same certificates are re-used across after a TLS relation is broken.)

juditnovak: ...or alternatively:

2. `certificates-relation-joined` 
     - We wipe out any existing certificates and request brand-new ones?

tests/integration/ha/test_network_cut.py
juditnovak: Yes, I'm getting back to do all changes at once, when all review is collected and discussions are finalized.
(In attempt to reduce context switch.)

src/events/tls.py
Mehdi-Bendriss: The flow should be the same as OpenSearch, where on:
1. `certificates-relation-broken`: We keep certificates if TLS is fully configured for the unit. 
2. `certificates-relation-created` (not `joined`): Request new certificates, regardless of the existence of certificates in the unit.

juditnovak: Guys at this point...

Please note that this ticket was meant to provide a _fix for broken HA tests_. That was the task we were meant to solve, with an **Acceptance Criteria** of having the HA network cut tests stable green. That has been completed, and any additional adjustments are outside of the scope of the task (that has been already extending for 7 weeks).

In case we are not happy with the TLS workflow that we have set up for Opensearch Dashboards, then we should re-engineer that workflow separately.
'Re-engineering the TLS workflow' is out of scope of the task of 'Fixing and stabilizing the HA network cut tests'.

Especially since this task has been extending 
 - for 7 weeks
 - has completed its purpose
 - was not meant to address internal re-design further than HA network cut tests stabilization required

 ...furthermore...
 
- the proposed changes are major to the OSD TLS workflow
- leaves important additional questions open (do we never delete outdated certificates then? or we delete them when `certificate-available` so we leave a new `csr` co-exist with old certificates...? if the "leftover" certs expire we just ignore? etc.
- requires additional testing (i.e. on broken TLS connection the service is still available -- new behavior)

I would be highly uncomfortable squeezing all the above in this task.
 - without the testing above code quality can't be assured
    - typically: no control if the new behavior may be broken in the future by additional code changes or bugs
 - this code change would completely loose it's consize and self-consistent nature (focusing on fixing a broken behavior)
    - typically contra-productive for rollbacks, or when browsing commits history
 
 Is there any particular issue adding this change as a separate task, in a dedicated PR, self-containing, with corresponding testing provided?
 
 (As a Developer I find it important to have consize, self-confaining tasks with clear Acceptance Criterias achieved, as it helps to see and demonstrate progress together with keeping up a good, dynamic moral. Furthermore I'm very uncomfortable not to have solid testing for expected behaviors of the code I'm providing ensuring the behavior is not only available at the time of the change but also preserved by future code changes.)

reneradoi: Hey @juditnovak, I agree with you and think that it would be ok to tackle this in a separate task. Like I wrote earlier: it's absolutely fine for me when you just put it into an issue or "enhancement request" for later decision.

Maybe we can agree on that?

juditnovak: To summarize the MM communication on this thread.

- @Mehdi-Bendriss was keen on having https://github.com/canonical/opensearch-dashboards-operator/pull/82#discussion_r1723900454 implemented withing this scope (he requested minimal code changes, no integration tests)
- During implementation the change revealed as **breaking*** compared to previous behavior:
   * we used to switch between HTTS/HTTP via the presence of a TLS connection
   * implementing this change we will have no option to switch back to HTTP

(Integration tests added on a voluntary bases, highlighting the new behavior)

tests/integration/ha/test_network_cut.py
juditnovak: Done.

reneradoi (approved):

Mehdi-Bendriss (approved):

PR #86, opened 2024-07-04
[DPE-4828] Move to opensearch-client name

### LINKED ISSUE #85: Adopt the same pattern for `opensearch-clients` endpoint names
We are currently diverging between opensearch and opensearch-dashboard relation name conventions. We should adopt "opensearch-client" as its client name.

phvalguima:
Get the same endpoint name for client: `opensearch-client`.

Closes #85


Mehdi-Bendriss (approved):

juditnovak (approved):
Awesome, thank you Pedro!!!!!!

PR #91, opened 2024-07-08
[DPE-4226] Health check against opesearch

juditnovak:
Currently there is no healthcheck on ODS, thus even if the service is down, the charm appears as active.

This PR introduces a check on the OSD Rest API `/api/status` page, determining the overall service status.

In addition, the service status is also altered in case all units may be in a particular status at the same time.

### DIFF
--- a/charm_version
+++ b/charm_version
@@ -1,1 +1,1 @@
-1.0+91f8440-dirty+91f8440-dirty+91f8440-dirty+91f8440-dirty+91f8440-dirty+91f8440-dirty+91f8440-dirty
+1.0
--- a/charm_version
+++ b/charm_version
@@ -1,1 +1,1 @@
-1.0+69db8cf+69db8cf-dirty+69db8cf-dirty+69db8cf-dirty+69db8cf-dirty+d0f9b21
+1.0
--- a/charm_version
+++ b/charm_version
@@ -1,1 +1,1 @@
-1.0
+1.0+91f8440-dirty+91f8440-dirty+91f8440-dirty+91f8440-dirty+91f8440-dirty+91f8440-dirty+91f8440-dirty
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -33,7 +33,9 @@ python = "^3.10"
 tenacity = "^8.4.2"
 pure-sasl = "^0.6.2"
 cosl = "^0.0.12"
-ops = "^2.13.0"
+pydantic = "^1.10.17"
+pyyaml = "^6.0.1"
+poetry-plugin-export = "^1.8.0"
 
 # TODO: clean any of the notes below and their deps.
 [tool.poetry.group.charm-libs.dependencies]
@@ -78,12 +80,18 @@ pep8-naming = "^0.14.1"
 codespell = "^2.2.6"
 pyright = "^1.1.318"
 typing-extensions = "^4.9.0"
+requests = "^2.32.3"
+ops = "^2.13.0"
+
+[tool.poetry.group.unit]
+optional = true
 
 [tool.poetry.group.unit.dependencies]
 pytest = "^8.2.2"
 coverage = {extras = ["toml"], version = "^7.5.1"}
 pytest-mock = "^3.11.1"
 pyyaml = "^6.0.1"
+responses = "^0.25.3"
 
 [tool.poetry.group.integration.dependencies]
 pytest = "^8.2.2"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -43,7 +43,7 @@ poetry-plugin-export = "^1.8.0"
 ops = "^2.13.0"
 poetry-core = "^1.9.0"
 # tls_certificates_interface/v3/tls_certificates.py
-cryptography = "*"
+cryptography = "42.0.8"
 jsonschema = "*"
 # pinning to avoid: https://github.com/canonical/charmcraft/issues/1722
 # We should unpin it once we have rustc 1.76+ available at build time
--- a/src/charm.py
+++ b/src/charm.py
@@ -136,14 +136,12 @@ class OpensearchDasboardsCharm(CharmBase):
     def reconcile(self, event: EventBase) -> None:
         """Generic handler for all 'something changed, update' events across all relations."""
 
-        outdated_status = []
-
         # 1. Block until peer relation is set
         if not self.state.peer_relation:
             self.unit.status = WaitingStatus(MSG_WAITING_FOR_PEER)
             return
-        else:
-            outdated_status.append(MSG_WAITING_FOR_PEER)
+
+        outdated_status = [MSG_WAITING_FOR_PEER]
 
         # attempt startup of server
         if not self.state.unit_server.started:
--- a/src/charm.py
+++ b/src/charm.py
@@ -7,28 +7,30 @@
 import logging
 import time
 
-# from events.provider import ProviderEvents
 from charms.grafana_agent.v0.cos_agent import COSAgentProvider
 from charms.rolling_ops.v0.rollingops import RollingOpsManager
 from ops.charm import CharmBase, InstallEvent, SecretChangedEvent
 from ops.framework import EventBase
 from ops.main import main
-from ops.model import BlockedStatus, MaintenanceStatus, WaitingStatus
+from ops.model import ActiveStatus, BlockedStatus, MaintenanceStatus, WaitingStatus
 
 from core.cluster import ClusterState
 from events.requirer import RequirerEvents
 from events.tls import TLSEvents
 from events.upgrade import ODUpgradeEvents, OpensearchDashboardsDependencyModel
-from helpers import clear_status
+from helpers import clear_global_status, clear_status, set_global_status
 from literals import (
     CHARM_KEY,
     COS_PORT,
     COS_RELATION_NAME,
     DEPENDENCIES,
+    MSG_DB_DOWN,
     MSG_DB_MISSING,
+    MSG_INCOMPATIBLE_UPGRADE,
     MSG_INSTALLING,
     MSG_STARTING,
     MSG_STARTING_SERVER,
+    MSG_STATUS,
     MSG_TLS_CONFIG,
     MSG_WAITING_FOR_PEER,
     PEER,
@@ -36,8 +38,11 @@ from literals import (
     SERVER_PORT,
     SUBSTRATE,
 )
+from managers.api import APIManager
 from managers.config import ConfigManager
+from managers.health import HealthManager
 from managers.tls import TLSManager
+from managers.upgrade import UpgradeManager
 from workload import ODWorkload
 
 logger = logging.getLogger(__name__)
@@ -56,12 +61,8 @@ class OpensearchDasboardsCharm(CharmBase):
 
         self.tls_events = TLSEvents(self)
         self.requirer_events = RequirerEvents(self)
-        self.upgrade_events = ODUpgradeEvents(
-            self,
-            dependency_model=OpensearchDashboardsDependencyModel(
-                **DEPENDENCIES  # pyright: ignore[reportGeneralTypeIssues]
-            ),
-        )
+        dependency_model = OpensearchDashboardsDependencyModel(**DEPENDENCIES)
+        self.upgrade_events = ODUpgradeEvents(self, dependency_model=dependency_model)
 
         # --- MANAGERS ---
 
@@ -70,6 +71,18 @@ class OpensearchDasboardsCharm(CharmBase):
         )
         self.config_manager = ConfigManager(
             state=self.state, workload=self.workload, substrate=SUBSTRATE, config=self.config
+        )
+        self.api_manager = APIManager(
+            state=self.state, workload=self.workload, substrate=SUBSTRATE
+        )
+        self.health_manager = HealthManager(
+            state=self.state, workload=self.workload, substrate=SUBSTRATE
+        )
+        self.upgrade_manager = UpgradeManager(
+            state=self.state,
+            workload=self.workload,
+            substrate=SUBSTRATE,
+            dependency_model=dependency_model,
         )
 
         # --- LIB EVENT HANDLERS ---
@@ -123,9 +136,15 @@ class OpensearchDasboardsCharm(CharmBase):
     def reconcile(self, event: EventBase) -> None:
         """Generic handler for all 'something changed, update' events across all relations."""
 
+        outdated_status = []
+
         # not all methods called
         if not self.state.peer_relation:
+            # Block until peer relation is set
+            self.unit.status = WaitingStatus(MSG_WAITING_FOR_PEER)
             return
+        else:
+            outdated_status.append(MSG_WAITING_FOR_PEER)
 
         # attempt startup of server
         if not self.state.unit_server.started:
@@ -135,11 +154,31 @@ class OpensearchDasboardsCharm(CharmBase):
         if getattr(event, "departing_unit", None) == self.unit:
             return
 
-        outdated_status = []
+        # Restart on config change
+        if (
+            self.config_manager.config_changed()
+            and self.state.unit_server.started
+            # and self.upgrade_events.idle
+        ):
+            self.on[f"{self.restart.name}"].acquire_lock.emit()
+            # No point in setting any status -- would be wiped out by rollingops after the restert
+            return
+
         # Maintain the correct app status
-        if self.unit.is_leader():
-            if self.state.opensearch_server:
-                outdated_status.append(MSG_DB_MISSING)
+        # No further actions below but only status settings
+
+        # Block until Opensearch is available and it's a compatible version
+        if self.state.opensearch_server:
+            outdated_status.append(MSG_DB_MISSING)
+        else:
+            set_global_status(self, BlockedStatus(MSG_DB_MISSING))
+            return
+
+        if self.upgrade_manager.version_compatible():
+            outdated_status.append(MSG_INCOMPATIBLE_UPGRADE)
+        else:
+            set_global_status(self, BlockedStatus(MSG_INCOMPATIBLE_UPGRADE))
+            return
 
         # Maintain the correct unit status
 
@@ -149,18 +188,32 @@ class OpensearchDasboardsCharm(CharmBase):
                 outdated_status.append(MSG_TLS_CONFIG)
             else:
                 self.unit.status = MaintenanceStatus(MSG_TLS_CONFIG)
+                return
+        else:
+            outdated_status.append(MSG_TLS_CONFIG)
 
-        # Restart on config change
-        if (
-            self.config_manager.config_changed()
-            and self.state.unit_server.started
-            # and self.upgrade_events.idle
-        ):
-            self.on[f"{self.restart.name}"].acquire_lock.emit()
+        # Regular health-check
+        if self.health_manager.opensearch_ok():
+            outdated_status.append(MSG_DB_DOWN)
+        else:
+            set_global_status(self, BlockedStatus(MSG_DB_DOWN))
+            return
+
+        healthy, msg = self.health_manager.healthy()
+
+        if not healthy:
+            self.unit.status = BlockedStatus(msg)
+            return
+
+        if msg:
+            self.unit.status = WaitingStatus(msg)
+            return
+        else:
+            outdated_status += MSG_STATUS
 
         # Clear all possible irrelevant statuses
         for status in outdated_status:
-            clear_status(self.unit, status)
+            clear_global_status(self, status)
 
     def _on_secret_changed(self, event: SecretChangedEvent):
         """Reconfigure services on a secret changed event."""
@@ -217,6 +270,7 @@ class OpensearchDasboardsCharm(CharmBase):
             time.sleep(5)
 
         clear_status(self.unit, [MSG_STARTING, MSG_STARTING_SERVER])
+        self.on.update_status.emit()
 
     # --- CONVENIENCE METHODS ---
 
@@ -229,6 +283,7 @@ class OpensearchDasboardsCharm(CharmBase):
         self.config_manager.set_dashboard_properties()
 
         logger.debug("starting Opensearch Dashboards service")
+
         self.workload.start()
 
         # open port
--- a/src/charm.py
+++ b/src/charm.py
@@ -12,7 +12,7 @@ from charms.rolling_ops.v0.rollingops import RollingOpsManager
 from ops.charm import CharmBase, InstallEvent, SecretChangedEvent
 from ops.framework import EventBase
 from ops.main import main
-from ops.model import ActiveStatus, BlockedStatus, MaintenanceStatus, WaitingStatus
+from ops.model import BlockedStatus, MaintenanceStatus, WaitingStatus
 
 from core.cluster import ClusterState
 from events.requirer import RequirerEvents
@@ -24,14 +24,14 @@ from literals import (
     COS_PORT,
     COS_RELATION_NAME,
     DEPENDENCIES,
-    MSG_DB_DOWN,
-    MSG_DB_MISSING,
+    MSG_APP_STATUS,
     MSG_INCOMPATIBLE_UPGRADE,
     MSG_INSTALLING,
     MSG_STARTING,
     MSG_STARTING_SERVER,
-    MSG_STATUS,
+    MSG_STATUS_DB_MISSING,
     MSG_TLS_CONFIG,
+    MSG_UNIT_STATUS,
     MSG_WAITING_FOR_PEER,
     PEER,
     RESTART_TIMEOUT,
@@ -138,9 +138,8 @@ class OpensearchDasboardsCharm(CharmBase):
 
         outdated_status = []
 
-        # not all methods called
+        # 1. Block until peer relation is set
         if not self.state.peer_relation:
-            # Block until peer relation is set
             self.unit.status = WaitingStatus(MSG_WAITING_FOR_PEER)
             return
         else:
@@ -154,24 +153,24 @@ class OpensearchDasboardsCharm(CharmBase):
         if getattr(event, "departing_unit", None) == self.unit:
             return
 
-        # Restart on config change
+        # 2. Restart on config change
         if (
             self.config_manager.config_changed()
             and self.state.unit_server.started
-            # and self.upgrade_events.idle
+            and self.upgrade_events.idle
         ):
             self.on[f"{self.restart.name}"].acquire_lock.emit()
             # No point in setting any status -- would be wiped out by rollingops after the restert
             return
 
-        # Maintain the correct app status
+        # 3. Maintain the correct app status
         # No further actions below but only status settings
 
         # Block until Opensearch is available and it's a compatible version
         if self.state.opensearch_server:
-            outdated_status.append(MSG_DB_MISSING)
+            outdated_status.append(MSG_STATUS_DB_MISSING)
         else:
-            set_global_status(self, BlockedStatus(MSG_DB_MISSING))
+            set_global_status(self, BlockedStatus(MSG_STATUS_DB_MISSING))
             return
 
         if self.upgrade_manager.version_compatible():
@@ -193,23 +192,26 @@ class OpensearchDasboardsCharm(CharmBase):
             outdated_status.append(MSG_TLS_CONFIG)
 
         # Regular health-check
-        if self.health_manager.opensearch_ok():
-            outdated_status.append(MSG_DB_DOWN)
-        else:
-            set_global_status(self, BlockedStatus(MSG_DB_DOWN))
-            return
-
-        healthy, msg = self.health_manager.healthy()
-
-        if not healthy:
-            self.unit.status = BlockedStatus(msg)
-            return
-
-        if msg:
-            self.unit.status = WaitingStatus(msg)
+        # Checks that may modify the 'app' state as well
+        app_healthy, app_msg = self.health_manager.app_healthy()
+        if not app_healthy:
+            set_global_status(self, BlockedStatus(app_msg))
             return
         else:
-            outdated_status += MSG_STATUS
+            outdated_status += MSG_APP_STATUS
+
+        # Checks purely on unit level
+        unit_healthy, unit_msg = self.health_manager.unit_healthy()
+
+        if not unit_healthy:
+            self.unit.status = BlockedStatus(unit_msg)
+            return
+
+        if unit_msg:
+            self.unit.status = WaitingStatus(unit_msg)
+            return
+        else:
+            outdated_status += MSG_UNIT_STATUS
 
         # Clear all possible irrelevant statuses
         for status in outdated_status:
@@ -297,7 +299,7 @@ class OpensearchDasboardsCharm(CharmBase):
         clear_status(self.unit, MSG_STARTING_SERVER)
 
         if self.unit.is_leader() and not self.state.opensearch_server:
-            self.app.status = BlockedStatus(MSG_DB_MISSING)
+            self.app.status = BlockedStatus(MSG_STATUS_DB_MISSING)
 
     def _scrape_config(self) -> list[dict]:
         """Generates the scrape config as needed."""
--- a/src/core/cluster.py
+++ b/src/core/cluster.py
@@ -59,10 +59,7 @@ class ClusterState(Object):
     @property
     def opensearch_relation(self) -> Relation | None:
         """The Opensearch Server relation."""
-        try:
-            return self.model.get_relation(OPENSEARCH_REL_NAME)
-        except KeyError:
-            return None
+        return self.model.get_relation(OPENSEARCH_REL_NAME)
 
     @property
     def tls_relation(self) -> Relation | None:
--- a/src/core/cluster.py
+++ b/src/core/cluster.py
@@ -32,7 +32,7 @@ class ClusterState(Object):
     """Collection of global cluster state for Framework/Object."""
 
     def __init__(self, charm: Framework | Object, substrate: SUBSTRATES):
-        super().__init__(parent=charm, key="charm_state")
+        super().__init__(parent=charm, key="osd_charm_state")
         self.substrate: SUBSTRATES = substrate
         self._servers_data = {}
 
@@ -59,7 +59,10 @@ class ClusterState(Object):
     @property
     def opensearch_relation(self) -> Relation | None:
         """The Opensearch Server relation."""
-        return self.model.get_relation(OPENSEARCH_REL_NAME)
+        try:
+            return self.model.get_relation(OPENSEARCH_REL_NAME)
+        except KeyError:
+            return None
 
     @property
     def tls_relation(self) -> Relation | None:
--- a/src/core/models.py
+++ b/src/core/models.py
@@ -12,7 +12,7 @@ from charms.data_platform_libs.v0.data_interfaces import Data, DataDict
 from ops.model import Application, Relation, Unit
 from typing_extensions import override
 
-from literals import COS_USER
+from literals import SERVER_PORT
 
 logger = logging.getLogger(__name__)
 
@@ -166,16 +166,6 @@ class ODServer(StateBase):
         return self.relation_data.get("state", None) == "started"
 
     @property
-    def cos_user(self) -> str | None:
-        """The generated password for the client application."""
-        return COS_USER
-
-    @property
-    def cos_password(self) -> str | None:
-        """The generated password for the client application."""
-        return self.relation_data.get("monitor-password")
-
-    @property
     def password_rotated(self) -> bool:
         """Flag to check if the unit has rotated their internal passwords."""
         return bool(self.relation_data.get("password-rotated", None))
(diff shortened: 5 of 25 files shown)


src/managers/api.py
```diff
@@ -0,0 +1,97 @@
...
+            "method": method.upper(),
+            "url": full_url,
+            "headers": headers,
+        }
+
+        request_kwargs["data"] = json.dumps(payload)
+        request_kwargs["headers"] = headers
+
+        if not self.state.opensearch_server:
+            raise OSDAPIError("No Opensearch connection, can't query API (missing credentials).")
```
Mehdi-Bendriss, line 79: I think the message is not entirely accurate here, as this can happen when the opensearch server is down too, not just on "missing credentials“

src/managers/api.py
```diff
@@ -0,0 +1,97 @@
...
+                resp.raise_for_status()
+        except RequestException as e:
+            logger.error(f"Request {method} to {full_url} with payload: {payload} failed. \n{e}")
+            raise
+
+        return resp.json()
+
+    def service_status(self) -> dict[str, Any]:
+        """Query service status from the OSD API."""
+        return self.request(endpoint="status")
```
Mehdi-Bendriss, line 97: will this API call return a `200` or other success codes if the Dashboards server is up while OpenSearch is not?

src/managers/health.py
```diff
@@ -0,0 +1,63 @@
...
+    MSG_STATUS_UNKNOWN,
+    MSG_STATUS_WORKLOAD_DOWN,
+)
+from managers.api import APIManager
+
+logger = logging.getLogger(__name__)
+
+
+class HealthManager:
+    """Manager for handling Kafka machine health."""
```
Mehdi-Bendriss, line 27:
```suggestion
    """Manager for handling OpenSearch Dashboards machine health."""
```

src/managers/health.py
```diff
@@ -0,0 +1,63 @@
...
+            status_data = self.api_manager.service_status()
+        except HTTPError as err:
+            if err.response.status_code == 503:
+                return False, MSG_STATUS_UNAVAIL
+        except (ConnectionError, OSDAPIError):
+            return False, MSG_STATUS_UNAVAIL
+
+        if status_data["status"]["overall"]["state"] == "green":
+            return True, ""
+        elif status_data["status"]["overall"]["state"] == "yellow":
```
Mehdi-Bendriss, line 52: what does `yellow` or other colors refer to?

Mehdi-Bendriss (commented):
Thanks Judit! Looks good, I have a few questions

src/core/cluster.py
```diff
@@ -59,7 +59,10 @@ def peer_relation(self) -> Relation | None:
     @property
     def opensearch_relation(self) -> Relation | None:
         """The Opensearch Server relation."""
-        return self.model.get_relation(OPENSEARCH_REL_NAME)
+        try:
+            return self.model.get_relation(OPENSEARCH_REL_NAME)
+        except KeyError:
```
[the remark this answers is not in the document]
juditnovak: Yes, I encountered `KeyError`...

I'll give it a try to reproduce and then raise the bug. However until the bug is fixed, we should make sure that our code is stable, so we should leave the check.

src/managers/api.py
juditnovak: We have missing credentials since Opensearch is down. Opensearch is providing our credentials (i.e. `kibanaserver`). 

Without authentication (that comes from Opensearch) the API is unreachable.

I'll try to reflect this more clear in the message.

src/charm.py
```diff
@@ -133,9 +142,12 @@ def reconcile(self, event: EventBase) -> None:
...
         # Maintain the correct app status
-        if self.unit.is_leader():
-            if self.state.opensearch_server:
-                outdated_status.append(MSG_DB_MISSING)
+        if self.state.opensearch_server:
+            outdated_status.append(MSG_DB_MISSING)
+        else:
+            self.unit.status = BlockedStatus(MSG_DB_MISSING)
+            if self.unit.is_leader():
+                self.app.status = BlockedStatus(MSG_DB_MISSING)
```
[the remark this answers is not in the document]
juditnovak: Cool, I didn't know. I double-check and remove the extra code if it's indeed that simple :-)

src/managers/api.py
juditnovak: Dasboard server is down if Opensearch is down. So is the whole API.

Thus `self.request()` exception handlers take effect in that case.

src/managers/health.py
juditnovak: How I wish this was documented... :-( 

Based on Opensearch, I assumed there may be a `yellow` (degraded) state. As I assume development of this API went much alongside of development of Elastic.

Now I find this: https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/core/server/status/legacy_status.ts#L114-L169
Let me know if you would like us to "hardcode" more of this into the dashboards. I have no idea how much we can rely on these values as I find no docs (that could assure none of this changes across versions...)

src/core/models.py
```diff
@@ -261,3 +261,10 @@ def sans(self) -> dict[str, list[str]]:
             "sans_ip": [self.private_ip],
             "sans_dns": [self.hostname, self.fqdn],
         }
+
+    @property
+    def url(self) -> str:
+        """Service URL."""
+        if self.tls:
+            return f"https://{self.private_ip}:{SERVER_PORT}"
+        return f"http://{self.private_ip}:{SERVER_PORT}"
```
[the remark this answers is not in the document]
phvalguima: Not using TLS is an option?

src/managers/health.py
juditnovak: (Regarding testing: I'm feeling a bit lost how to "squeeze out" these statuses from OSD -- since to obvious anomalies, like Opensearch being down the OSD service itself also just goes down. Thus I'm unsure how to have integration tests or even manual testing for states of various colors. Of course unit testing is already there for 'yellow', and I could extend that in case we decide to handle the full granularity.)

Alternatively we can stick to a "rule of thumb" policy as the current one:
 - green - good
 - yellow - mark it in the message but don't change unit status to blocked
 - anything else - unit is blocked, we report error
 
(I assume this may safe enough to hold across version changes and so.)

src/charm.py
```diff
@@ -7,7 +7,7 @@
 import logging
 import time
 
-from charms.rolling_ops.v0.rollingops import RollingOpsManager
+from charms.rolling_ops.v0.rollingops import ActiveStatus, RollingOpsManager
```
phvalguima, line 10: This should not be imported from rollingops

src/managers/api.py
```diff
@@ -0,0 +1,97 @@
...
+        try:
+            with requests.Session() as s:
+                s.auth = (  # type: ignore [reportAttributeAccessIssue]
+                    self.state.opensearch_server.username,
+                    self.state.opensearch_server.password,
+                )
+                resp = s.request(**request_kwargs)
+                resp.raise_for_status()
+        except RequestException as e:
+            logger.error(f"Request {method} to {full_url} with payload: {payload} failed. \n{e}")
```
phvalguima, line 90: I suggest raising a specific exception here, so you later in your code, you can watch for some few only exceptions in order to cover the entire code:
```suggestion
            logger.error(f"Request {method} to {full_url} with payload: {payload} failed. \n{e}")
            raise ODSHttpException(
                response_text=ex.response.text,
                response_code=ex.response.status_code,
            )  # composed of status code error and any message
```

src/managers/api.py
```diff
@@ -0,0 +1,97 @@
...
+            raise
+
+        return resp.json()
```
phvalguima, line 93: Can we also return the status code, as tuple here? I believe resp always contains a `resp.status_code`

tests/integration/helpers.py
```diff
@@ -248,17 +293,22 @@ async def access_all_dashboards(
...
-            f"JUJU_MODEL={model_full_name} juju scp "
-            f"ubuntu@{unit}:/var/snap/opensearch-dashboards/current/etc/opensearch-dashboards/certificates/ca.pem ./",
-        ],
-    )
+    try:
+        output = subprocess.run(
+            [
+                "bash",
+                "-c",
+                f"JUJU_MODEL={model_full_name} juju scp "
```
phvalguima, line 304: You can do the same with [libjuju scp_{to|from}](https://github.com/juju/python-libjuju/blob/6c6d70d483937cec4404116c376b926106e05835/juju/unit.py#L313-L351)

phvalguima (changes requested):

src/charm.py
juditnovak: Indeed, it was a silent automatism, fixing.

src/managers/api.py
juditnovak: I thought of that option, and after some thinkgin/coding I purposefully chose that `self.request()` would return Python `requests` errors. (Generally I also rather prefer to have the approach you're describing.)

I quite like how Python `requests` groups its own exceptions under RequestException (much alongside the same consideration as what you are raising), and I prefer to distinguish later -- based on whther `self.requests()` returned a `ConnectionError` or a `HttpError` or an `SSLError` (all descendats of `RequestException`).

Typically if we want to further refine healthchecks or so -- we would like to distinguish between those, and not just hiding them (and reverse envineering them) from under a local exception's umbrella.

Alternatively of course we can also say that ANY of those result should in a `blocked` service... Yet we can't even devermine a charm status message in an elegant way if we do that.

src/managers/api.py
juditnovak: If my logic here is correct, it would be redundant. I mean we don't need it.
We ONLY return results in case 200.

`respose.raise_for_status()` is the last call within the `try` block.
In case the request was not successful, an exception is raised and bubbled up via `RequestsException` ('s descenant `HttpError`)

https://requests.readthedocs.io/en/latest/user/quickstart/#errors-and-exceptions

Pls correct me if I'm wrong (it could be the case :sweat_smile: )

tests/integration/helpers.py
juditnovak: I can give it a try (again. I recall I had an attempt to use it, and I came across missing options that can be directly passed to `scp` while not available when using the `libjuju` implementaiton of it)

src/managers/api.py
juditnovak: (Note that I'm purposefully using `repose.raise_for_status()` to have any HTTP errors bubbled up via `RequestException` for potential further analysis.)

src/managers/health.py
Mehdi-Bendriss: In that case, perhaps:
- green: good 
- yellow: `WaitingStatus(waiting for service health to improve...)`
- red: BlockedStatue

Mehdi-Bendriss: > unsure how to have integration tests or even manual testing for states of various colors

For green, it's clear. However for "red" or opensearch unreachable, I would have an integration test where, we:
1. juju ssh into all `opensearch`units
2. send a `SIGSTOP` signal to opensearch
3. ensure that all the opensearch nodes are indeed not online
4. wait for status of the dashboards to reflect the blocked state (opensearch server not reachable)
5. juju ssh into all opensearch units 
6. send a `SIGCONT` signal to opensearch
7. wait for the nodes to be online 
8. wait for status of the dashboards to be active 
9. check that units of the dashboards are online (http 200)  

For the opensearch part, you can refer to this HA test 
https://github.com/canonical/opensearch-operator/blob/main/tests/integration/ha/test_ha.py#L262-L264

juditnovak: I see your point, but the above does not test the `api/status` "red" state.

The above test --from Opensearch Dashboards service pespective -- is almost equal to the existing test, where we break the relation to Opensearch.

Opensearch Dashboards responds to both situations the same way. Service goes down. There is NO `api/status`. Not in any color.

I can implement the test if you still prefer, however in terms of the healthcheck it is redundant to the already existing healthcheck. As the healthcheck --in both cases-- finds that the API is down.

However I see a valid use case that is currently not covered -- which the test above is spotting on.
In addition to the charm relation check, we should check Opensearch status (via an Opensearch REST API query). This functionality is indeed missing from the current solution. And this is to be detected by the test.

How should we check on Opensearch status? Accessing all known endpoints if they are all down?

juditnovak: So to summarize the offline discussion:

 - indeed we're missing Opensearch status detection (Opensearch API status). To be added.
   - with the test above
 - another missing functionality was discovered in the meanwhile: version mismatch detecion. To be added.

src/managers/api.py
phvalguima: @juditnovak missing catching `RequestException` here

tests/integration/helpers.py
```diff
@@ -139,6 +140,50 @@ def restart_unit(model_full_name: str, unit: str) -> None:
...
+        host = f"[{host}]"
+
+    protocol = "http" if not https else "https"
+    url = f"{protocol}://{host}:5601/auth/login"
+    arguments = {"url": url}
+    if https:
+        arguments["verify"] = "./ca.pem"
+
+    try:
+        response = requests.get(**arguments)
```
phvalguima, line 164: Only connection errors? Why not SSL or HTTP errors as well?

src/managers/api.py
phvalguima: Ok, then please make it clear on the docstring of this method which exceptions should we expect and why

tests/integration/helpers.py
juditnovak: If it's an SSL error, then the dasboard is not unvailable, but has TLS misconfigured.
I can add a `return False` on that.
We do not `raise_for_status)` here, so I don't expect any HTTP errors.

src/managers/api.py
juditnovak: It's not missing. IT's by design.

This is the API Manager, implementing requests and a thing layer of calling specific REST API calls. IT's purposefully does no more logic but calling the API.

In the healthcheck module we are handling exceptions in case any API calls may not have returned expected results.

tests/integration/helpers.py
juditnovak: (NOTE: this function is meant to verify if the dashboard as a service is down. Misconfigured state doesn't count as down for us in this context. I can clarify in the docstring.)

src/managers/api.py
juditnovak: Note for the above: `SSLError` is a descendant of `ConnectionError`.

phvalguima: @juditnovak then, it must be explicit in the docstring this method will throw exceptions from `requests`. Otherwise, I may not do the check into what `self.requests` throw and then get some random exception at runtime...

tests/integration/helpers.py
```diff
@@ -139,6 +140,50 @@ def restart_unit(model_full_name: str, unit: str) -> None:
     )
 
 
+@retry(
+    stop=stop_after_attempt(3),
+    wait=wait_fixed(5),
+    retry_error_callback=lambda _: False,
+    retry=lambda x: x is False,
+)
+def dashboard_unavailable(host: str, https: bool = False) -> bool:
```
phvalguima, line 149: Can we have `dashboard_available` instead?  Otherwise we have double negatives here...

A normal check will be in this case:
```
if not dashboard_unavailable():
    # Dashboard is healthy
```

tests/integration/helpers.py
phvalguima: Check my comment above. Having the dashboard healthy is the "normal" scenario. Therefore, this method will largely be used as a double negation (i.e. I want to know when `not dashboard_unavailable()`).

In the case we invert to `dashboard_available()`, then it makes sense to catch all the `request` exceptions and return them as False, then returning:
`return 200 <= response.status and response.status < 300`

phvalguima: Currently, this method will return: `True, False, throw exceptions`. It means a simple check will always have to be covered with a `try/except` block unnecessarily...

tests/integration/helpers.py
```diff
@@ -139,6 +140,50 @@ def restart_unit(model_full_name: str, unit: str) -> None:
...
+        arguments["verify"] = "./ca.pem"
+
+    try:
+        response = requests.get(**arguments)
+    except ConnectionError:
+        return True
+    return response.status_code == 503
+
+
+def all_dashboards_unavailable(ops_test: OpsTest, https: bool = False) -> bool:
```
phvalguima, line 170: Likewise, can we have it as `all_dashboards_available`

phvalguima (commented):
The double negation in `*dashboard(s)_unavailable` is confusing. It did confuse me in my review, as I'd expect we try to capture all the exceptions, but given we are using this "non-available", then it is less clear.

tests/integration/helpers.py
juditnovak: No double negation. That would be a mis-use of this helper function. (I can clarify this in the docstring.)

We already have `access_dashboard()` and `access_all_dashboards()` functions used all across all tests (including the test module in and new test this PR). That is the one to test Dashboards full healthy availability. Should return `True` if Dashboads is in a perfect shape. `False` on partial availabilty.

The new method is to check that Dashboard is specifically down (fully unavailable), and should not return `True` otherwise. If the Dashboard is available to "any extent" it should be `False`.

I can clarify on docstings or naming terminology. But I what I have in mind is 2 very simple use `True`/`False` cases with no additional logic, answering these questions:
 1. "Is the Dashboards is up and all well?" (If anything is wrong, "up but not well", etc -> `False`)
 3. "Is Dasboards sevice totally down?" (If serivce is green/yellow/red, i.e. up to ANY extent -> `False`)

It may worth to highlight that Dashboards is an extremely senitive service, going down on any unpreferred condition :-)
IF Opensearch can't be contacted (let's say `kibanaserver` pw is wrong): Dashboards simpy does not come up. IT's completely unavailable. No "red" status. You just can't access the API at all.

This is the reason why I think it makes sense to have a specific test helper function dedicated to detect this particular scenario. 

NOTE: Exceptions shouldn't be thrown. Only `True`/`False`. Which exceptions do you expect to he thrown? I may have overlooked something. If there's any, that must be handled.

May this context better highlight the intention here? If yes,

tests/integration/helpers.py
juditnovak: We already do...

It's called `access_dashboard()` and `acceess_all_dashboards()`. Used across ALL test modules to check 4 healthy status (including the one in this PR).

More info in comment https://github.com/canonical/opensearch-dashboards-operator/pull/91#discussion_r1677578612

tests/integration/helpers.py
juditnovak: We already do.... See `access_all_dashboars()`.

phvalguima: @juditnovak what is the use case you need to know all dashboards are *unavailable*?

tests/integration/helpers.py
phvalguima: @juditnovak then what is this module adding here?

src/core/models.py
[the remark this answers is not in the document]
juditnovak: Yes, it is for future traefik integrstions.

Unlike in Opensearch, here we purposefully keep this option up.

src/managers/health.py
juditnovak: Updates: both added.

tests/integration/test_charm.py
```diff
@@ -124,134 +126,198 @@ async def test_dashboard_access_https(ops_test: OpsTest):
...
-            count_lines_with(
-                ops_test.model_full_name,
-                unit.name,
-                "/var/snap/opensearch-dashboards/common/var/log/opensearch-dashboards/opensearch_dashboards.log",
-                "debug",
-            )
-            == debug_lines
-        )
+    logger.info("Destroying and restoring the Opensearch cluster")
+    await destroy_cluster(ops_test, app=OPENSEARCH_APP_NAME)
```
juditnovak, line 309: Unfortunately this doesn't seem stable.
https://github.com/canonical/opensearch-dashboards-operator/actions/runs/10175100391/job/28142472214#step:28:326

So I'm thinking to cut this out and add back a scaleup once the node election is fixed.

src/managers/api.py
juditnovak: Docstrings are now representing correctly the exceptions that may be thrown.

Closing thread.

src/charm.py
juditnovak: Fixed.

src/managers/health.py
juditnovak: Fixed.

src/managers/api.py
juditnovak: Docstring added reflecting exceptions that may be thrown.

src/managers/api.py
juditnovak: `"Can't query OSD API, no Opensearch connection (i.e. no OSD credentials)."`

Is that any better? I if not: what would you suggest?

(Note that this is the `managers/api.py` module. Not a regular (health) check that should report on system/service status. In this workflow the direct blocker making the API request is that we have no credentials to use for the query.)

src/core/models.py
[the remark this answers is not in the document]
juditnovak: Done.

src/charm.py
```diff
@@ -149,18 +187,35 @@ def reconcile(self, event: EventBase) -> None:
...
+            outdated_status += MSG_APP_STATUS
 
-        # Restart on config change
-        if (
-            self.config_manager.config_changed()
-            and self.state.unit_server.started
-            # and self.upgrade_events.idle
-        ):
-            self.on[f"{self.restart.name}"].acquire_lock.emit()
+        # Checks purely on unit levelpurely on unit level
```
reneradoi, line 203: nit: That comment is duplicated somehow.

src/managers/health.py
```diff
@@ -0,0 +1,114 @@
...
+
+    def opensearch_ok(self) -> tuple[bool, str]:
+        """Verify if associated Opensearch service is up and running."""
+
+        if not self.state.opensearch_server or not self.state.opensearch_server.tls_ca:
+            return False, MSG_STATUS_DB_MISSING
+
+        if not os.path.isfile(OPENSEARCH_CA_FILE):
+            with open(OPENSEARCH_CA_FILE, "w") as text_file:
+                text_file.write(self.state.opensearch_server.tls_ca)
```
reneradoi, line 72: Please be aware that the CA of Opensearch might change (CA rotation), that means the file you have locally could be out of date. This would then lead to a verification error on the request.

I would recommend to get and write the file each time, not just if it's missing.

reneradoi (commented):
Hey Judit, I've added one comment regarding the CA file topic that we discussed in our session. Other than that it looks fine.

src/managers/health.py
juditnovak: Thank U so much for keeping this one in mind.

Indeed we need to check it each time fresh. Also I should use the file write method from `workload`. Thanks much!!!

juditnovak: All right. In fact, I just noticed that I'm already caching the Opensearch CA file from when the relation is joined.

And that one is updated everytime the Opensearch relation (data) has changed.

So we shoudl be good :-)

src/charm.py
juditnovak: Fixed.

reneradoi (dismissed):

tests/integration/helpers.py
```diff
@@ -399,6 +450,32 @@ async def get_application_hosts(ops_test: OpsTest, app_name: str, units: List[st
     return hosts
 
 
+async def check_full_status(
+    ops_test: OpsTest, app: str = APP_NAME, status: str = "active", status_msg: str | None = None
+) -> bool:
+    """Compare app and unit status against those requested in the parameters."""
+    status_data = await ops_test.model.get_status()  # noqa: F821
+
+    if not status_data.applications[app].status.status == status:
```
Mehdi-Bendriss, line 459: Can you confirm that `.status.status` is correct here?

Mehdi-Bendriss (commented):
Thanks Judit - I have a few minor comments.

src/charm.py
```diff
@@ -135,11 +153,31 @@ def reconcile(self, event: EventBase) -> None:
...
+
+        # 3. Maintain the correct app status
+        # No further actions below but only status settings
+
+        # Block until Opensearch is available and it's a compatible version
+        if self.state.opensearch_server:
+            outdated_status.append(MSG_STATUS_DB_MISSING)
+        else:
+            set_global_status(self, BlockedStatus(MSG_STATUS_DB_MISSING))
+            return
```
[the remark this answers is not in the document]
juditnovak: I was thinking about this myself, and at the end consciously decided not to go for such.

_"Explicit is better than implicit"_ 

I prefer to reduce the level of redirections an keep the code clear and straightforward. I agree that this this pattern is repetitive, however I still rather prefer it repetitive but clear rather than "yet another" "magical helpler function" that is calling other helper functions, and  the reader needs to seek and resolve this whole "tangled-up" workflow in other functions/files. 

Since there is no issue on readability or clarity regarding the current solution, (the only issue is that it's "too verbose/repetitive") I'd prefer to leave it as such -- at least until I may find a way I'm also feeling comofortable with.

src/managers/upgrade.py
```diff
@@ -0,0 +1,46 @@
...
+        self.state = state
+        self.workload = workload
+        self.substrate = substrate
+        self.dependency_model = dependency_model
+
+    def version_compatible(self) -> bool:
+        """Verify version compatibility with Opensearch."""
+        # When there's no Opensearch connection, we shouldn't report version mismatch
+        if not self.state.opensearch_server:
+            return True
```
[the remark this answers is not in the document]
juditnovak: 1. Linting won't let us get rid of this -- as in the next line I'm referring to a `self.state.opensearch_server` attribute.
2. We consider that there is no verison compatibility issue when there is no connection to Opensearch. To me this fully makes sense (like mulitplication with zero results in zero). It's a base definition/assumption.

I'm not sure if I got your argumetnation correctly. 

> if the server is down

Do you mean the Openserach server? The above check is independent from the Openserach server being up or down. The above check is stricly looking at the existence of an Opensearch relation.

tests/integration/helpers.py
```diff
@@ -165,6 +167,50 @@ async def access_all_prometheus_exporters(ops_test: OpsTest) -> bool:
...
+    url = f"{protocol}://{host}:5601/auth/login"
+    arguments = {"url": url}
+    if https:
+        arguments["verify"] = "./ca.pem"
+
+    try:
+        response = requests.get(**arguments)
+    except ConnectionError:
+        return True
+    return response.status_code == 503
```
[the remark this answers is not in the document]
juditnovak: Because we are ONLY interested in the Dashboard being fully unavailable.

Which is either a `ConnectionError` or a `503` status returned.

src/managers/upgrade.py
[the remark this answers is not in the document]
juditnovak: See https://github.com/canonical/opensearch-dashboards-operator/pull/91/files/36eba6f1549b584f11ec960840e8c3500407d5b7#diff-9dc137d1679e723fa586c453222bc0a25cfa1dccf43430336148c13fd177c382L46

I think we should keep the scope of our PRs somewhat consize, not to change an upgrade strategy in a "Healthcheck/Status"-related modification.

tests/integration/helpers.py
juditnovak: image

Yes, it is correct.

src/managers/upgrade.py
```diff
@@ -0,0 +1,46 @@
...
+
+        srv_version_actual = self.state.opensearch_server.version
+        if not srv_version_actual:
+            return False
+
+        srv_version_required = self.dependency_model.osd_upstream.dependencies["opensearch"]
+        major_actual, minor_actual = srv_version_actual.split(".")[:2]
+        major_required, minor_required = srv_version_required.split(".")[:2]
```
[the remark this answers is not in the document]
juditnovak: I see your point.

However I'm not feeling comfortable changing the upgrade workflow within the scope of the  in the healthcheck modifications. Note that the current strategy is what we've agreed on when implementing upgrades. The logic just got moved from one module to another.

https://github.com/canonical/opensearch-dashboards-operator/pull/91/files/36eba6f1549b584f11ec960840e8c3500407d5b7#diff-9dc137d1679e723fa586c453222bc0a25cfa1dccf43430336148c13fd177c382L46

The PR is getting high in volume (even excluding 700 lines of poetry.lock). Once we did not think of handling patch version when implementing upgrades, then I think we should address it in a dedicated space (tests included, etc.). Not in a 1000+ lines PR of a different scope.

src/core/cluster.py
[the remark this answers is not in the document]
juditnovak: I tried, but so far I can't reproduce the issue. However I remember that I added this code for a reason. It would be difficult to report anything to `operator` w/o more precision.

Therefore we have 3 choices.

1. We remove the check and add it back once the error was observed again. (I don't know the risks for this solution with GE around...)
2. We keep the --otherwise harmless-- check, perhaps with logging added. This way we can track if the case occurs, and we're protected in case it may.
3. We open a separate ticket on investigating the issue (as it IS a unit of work, more than 30 mins)

I'm sorry to have let this one go (I didn't realize at the moment that it was a bug).

However, in my defense I'm spending non-negligable time lately on creating (descriptive) bug reports of Opensearch, Juju, Multipass, etc. (https://bugs.launchpad.net/juju/+bug/2074195, https://github.com/canonical/multipass/issues/3603, https://github.com/canonical/opensearch-operator/issues?q=is%3Aissue+author%3Ajuditnovak - more than 2 of those during this specific PR). Sometimes it feels that my own tasks are hardly progressing while providing (detailed, quality) bug reports for other tools :-(

src/managers/upgrade.py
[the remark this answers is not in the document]
juditnovak: Agreed on a live conversation: we're adjusting the upgrades strategy in a separate ticket.

https://warthogs.atlassian.net/browse/DPE-5070

src/core/cluster.py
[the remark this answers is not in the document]
juditnovak: Agreed live: now we remove the check and in case we come across it again we'll investigate (open a bug, add a workaround, etc.)

src/managers/api.py
phvalguima: @juditnovak all good for me. Leaving this thread open as it was originally raised by @Mehdi-Bendriss

phvalguima (approved):

tests/integration/test_charm.py
juditnovak: @Mehdi-Bendriss Pls let me know your thoughts on this. I don't have a better idea to get green pipelines (with at least partial test coverage)

src/managers/upgrade.py
```diff
@@ -0,0 +1,46 @@
...
+        if major_actual > major_required or minor_actual > minor_required:
+            return False
+        return True
```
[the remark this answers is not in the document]
juditnovak: Should be `<=`, adding it that way :-)

src/charm.py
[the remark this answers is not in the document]
juditnovak: See [related Matrix discussion](https://matrix.to/#/!eNCNzcteYUDDYpStsu:ubuntu.com/$EACSLpQgBn0fXIiceimpW7u3dM70xao_Qsw_2Lf_fRQ?via=ubuntu.com&via=matrix.org) and the [Juju bug](https://bugs.launchpad.net/juju/+bug/2074195) opened as a result.

What we do as a workaround for now, is to set `app` status whenever an issue (typically: Opensearch unavailability) is detected that's most likely impacting the whole cluster.

pyproject.toml
```diff
@@ -33,15 +33,17 @@ python = "^3.10"
...
+poetry-plugin-export = "^1.8.0"
 
 # TODO: clean any of the notes below and their deps.
 [tool.poetry.group.charm-libs.dependencies]
 # data_platform_libs/v0/data_interfaces.py
 ops = "^2.13.0"
 poetry-core = "^1.9.0"
 # tls_certificates_interface/v3/tls_certificates.py
-cryptography = "*"
+cryptography = "42.0.8"
```
juditnovak, line 46: Pinning it down to avoid
image

Providing a fix in https://github.com/canonical/opensearch-dashboards-operator/pull/82

src/managers/upgrade.py
```diff
@@ -0,0 +1,46 @@
...
+from charms.data_platform_libs.v0.upgrade import DependencyModel
+
+from core.cluster import SUBSTRATES, ClusterState
+from core.workload import WorkloadBase
+
+logger = logging.getLogger(__name__)
+
+
+class UpgradeManager:
+    """Manager for building necessary files for Java TLS auth."""
```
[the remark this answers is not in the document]
juditnovak: Done.

src/managers/upgrade.py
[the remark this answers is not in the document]
Mehdi-Bendriss: 1. It is fine. We can exclude this warning if needed.
2. Should the dashboards charm not be able to upgrade unless related to OpenSearch?

tests/integration/test_charm.py
```diff
@@ -261,3 +262,65 @@ async def test_log_level_change(ops_test: OpsTest):
...
+        apps=[APP_NAME, OPENSEARCH_APP_NAME], status="active", timeout=1000
+    )
+    assert ops_test.model.applications[APP_NAME].status == "active"
+    assert all(
+        unit.workload_status == "active" for unit in ops_test.model.applications[APP_NAME].units
+    )
+
+    logger.info("Checking if Dashboards is available again")
+    opensearch_relation = get_relations(ops_test, OPENSEARCH_RELATION_NAME)[0]
+    assert access_all_dashboards(ops_test, opensearch_relation, https=True)
```
Mehdi-Bendriss, line 298: 1. I believe we should pass the `opensearch_relation`'s ID not the object 
2. 
```suggestion
    assert await access_all_dashboards(ops_test, opensearch_relation, https=True)
```

tests/integration/test_charm.py
Mehdi-Bendriss: Please comment it for now and create a new ticket, and add a TODO comment to un-comment once the "ticket id" is resolved.

tests/integration/test_charm.py
```diff
@@ -261,3 +262,65 @@ async def test_log_level_change(ops_test: OpsTest):
...
+        await ops_test.model.wait_for_idle(apps=[OPENSEARCH_APP_NAME], status="blocked")
+
+    await ops_test.model.integrate(APP_NAME, OPENSEARCH_APP_NAME)
+    await ops_test.model.wait_for_idle(
+        apps=[APP_NAME, OPENSEARCH_APP_NAME], status="active", timeout=1000
+    )
+
+    logger.info("Checking if Dashboards is available again")
+    opensearch_relation = get_relations(ops_test, OPENSEARCH_RELATION_NAME)[0]
+    assert access_all_dashboards(ops_test, opensearch_relation, https=True)
```
Mehdi-Bendriss, line 326: 1. I believe we should pass the `opensearch_relation`'s ID not the object 
2. 
```suggestion
    assert await access_all_dashboards(ops_test, opensearch_relation, https=True)
```

Mehdi-Bendriss (commented):
Thanks Judit - I left a couple of comments on the tests

tests/integration/test_charm.py
juditnovak: You are totally right, thanks for noticing!!!!

juditnovak:
@Mehdi-Bendriss 
Sorry for lint/unittest failures. Pipelines have just revealed a [weak spot](https://github.com/canonical/opensearch-dashboards-operator/actions/runs/10287506585/job/28471711880#step:29:551) in the code (after we all thought it was getting "crystalized"). A fix was added immediately but tests weren't adjusted yet accordingly. (Note: failures were harmless, not signaling a wrong behavior.)

src/managers/upgrade.py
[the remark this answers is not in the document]
juditnovak: Done.

src/managers/upgrade.py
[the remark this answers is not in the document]
juditnovak: Dashboards charm is `blocked` (as a 'charm') if not related to Opensearch.
As a 'service' it's literally unavailable. The service doesn't come up.

I assume in real life most OSD Rolling Upgrades would happen
 - when the charm is in an `active` state
 - the service is healthy
 
Yet.

The current checks also allow for OSD upgrades when not related to Opensearch. Technically possible. After all, it's a valid scenario (as long as upgrades libs allow for an upgrade from a `blocked` status).

In this case potential version incompatibility will be detected during the relation establishing (as it will trigger `reconcile()` right after the reboot, and new healthchecks will detect the mis-match.).

Mehdi-Bendriss (approved):

charm_version
```diff
@@ -1 +1 @@
-1.0+69db8cf+69db8cf-dirty+69db8cf-dirty+69db8cf-dirty+69db8cf-dirty+d0f9b21
+1.0+91f8440-dirty+91f8440-dirty+91f8440-dirty+91f8440-dirty+91f8440-dirty+91f8440-dirty+91f8440-dirty
```
Mehdi-Bendriss, line 1: Please set this to `1.0`

juditnovak: Good point, thanks for bringing it up.
Do we want to keep it 1.0 permanently?

juditnovak: Done

Mehdi-Bendriss (dismissed):

phvalguima (dismissed):

phvalguima (dismissed):

Mehdi-Bendriss (approved):

phvalguima (approved):

Mehdi-Bendriss (approved):

PR #93, opened 2024-07-10
[DPE-4763] COS basics

juditnovak:
Dashboards COS integration for:
 - Prometheus
 - Grafana
 - Loki
 - Alerting

Tested locally, currently only relation data correctness is tested here.

Prometheus:

Screenshot from 2024-07-12 19-44-33

Grafana:

Screenshot from 2024-07-19 01-02-23

Loki:

Screenshot from 2024-07-19 00-06-52

Alerts:
Screenshot from 2024-07-19 01-50-49

Screenshot from 2024-07-19 01-51-55

Screenshot from 2024-07-19 02-01-24

Screenshot from 2024-07-19 02-01-41


actions.yaml
```diff
@@ -11,18 +11,6 @@ get-password:
 pre-upgrade-check:
   description: Run necessary pre-upgrade checks before executing a charm upgrade.
 
-set-password:
```
phvalguima, line 14: No more set-password is possible?

phvalguima, line 14: No more set-password is possible?

src/events/password_actions.py
```diff
@@ -1,73 +0,0 @@
-#!/usr/bin/env python3
-# Copyright 2023 Canonical Ltd.
-# See LICENSE file for licensing details.
-
-"""Event handlers for password-related Juju Actions."""
```
phvalguima, line 5: why going away?

tests/integration/test_charm.py
```diff
@@ -209,27 +216,32 @@ async def test_dashboard_password_rotation(ops_test: OpsTest):
...
-    password = await get_user_password(ops_test, user)
-    assert len(password) == 32
-
-    leader_num = await get_leader_id(ops_test)
-
-    # Change both passwords
-    result = await set_password(ops_test, username=user, num_unit=leader_num)
-    assert f"{user}-password" in result.keys()
-
+async def test_cos_relations(ops_test: OpsTest):
```
phvalguima, line 219: Missing the same test as we have on `opensearch`, where you `curl` directly the `/metrics` endpoints and see if there is data coming out.

phvalguima (commented):
It is not clear for me why a COS PR is removing the `set-password`, or why password-related code is going away.

actions.yaml
juditnovak: No, we actually never could set a password (so far).

We can:
 - either extend Opensearch Dashboards to allow for setting Opensearch user passwords
 - or stick to the current approach were user passwords are to be set directly on Opensearch

(Note that Dashboards is only using Opensearch users. It doesn't have any users on its own.)

src/events/password_actions.py
juditnovak: This has never done anything else but setting a secret that the charm associated with a "fiktive" user that never existed.

tests/integration/test_charm.py
juditnovak: Note: this request introduces a dependency to the PR.

We can only implement that test after:
 - we have the snap deploying the exporter service
 - that snap has been released (note that we have major issues around dashboards snap releases)
 - that snap has been taken into the Dashboards code.

Fine with me, we can delay this code to that point. Alternatively we can have quick small PRs avoiding too much to pile up for last minute.

juditnovak: Linked it against https://warthogs.atlassian.net/browse/DPE-4879 being 'blocked by', and moved the ticket to a BLOCKED state

actions.yaml
phvalguima: Yes, makes sense to remove it then

phvalguima (dismissed):
Nice! LGTM

src/charm.py
```diff
@@ -78,6 +76,17 @@ def __init__(self, *args):
...
 
+        # --- COS ---
+        self.cos_integration = COSAgentProvider(
+            self,
+            relation_name=COS_RELATION_NAME,
+            metrics_endpoints=[],
+            scrape_configs=self._scrape_config,
+            refresh_events=[self.on.config_changed],
+            # metrics_rules_dir="./src/alert_rules/prometheus",
+            # log_slots=["opensearch:logs"],
```
Mehdi-Bendriss, line 87: With the patch you brough on the snap, this slot can be re set now I think `opensearch-dashboards:log`

src/charm.py
```diff
@@ -218,11 +222,6 @@ def _restart(self, event: EventBase) -> None:
 
     def init_server(self):
         """Calls startup functions for server start."""
-        # don't run if leader has not yet created passwords
-        if not self.state.cluster.internal_user_credentials:
```
Mehdi-Bendriss, line 222: Wouldn't this run and attempt to start the server even prior to having a relation with `opensearch` (and receiving kiabanserver)?

src/charm.py
```diff
@@ -245,6 +244,19 @@ def init_server(self):
...
+    def _scrape_config(self) -> list[dict]:
+        """Generates the scrape config as needed."""
+        return [
+            {
+                "metrics_path": "/metrics",
+                "static_configs": [
+                    {"targets": [f"{self.state.unit_server.private_ip}:{COS_PORT}"]}
+                ],
+                # "tls_config": {"ca": self.state.unit_server.ca},
+                "scheme": "http",
```
Mehdi-Bendriss, line 256: Should we add a conditional statement here? when TLS is enabled vs when it's not?

Mehdi-Bendriss (commented):
Thanks Judit - it looks good. I only have a couple of questions to clarify

src/charm.py
juditnovak: Yes, I'm adding those changes all together once the snap version van be increased

src/charm.py
juditnovak: `kibanasever` was never returned by `self.state.cluster.internal_user_credentials`.

It was only `CHARM_USERS` that contained `monitor` only. (Note `monitor` user never existed, we just "pretended" it's there by setting a `monitor-password` for it. Has never been created for real.)

src/charm.py
juditnovak: `kibana-exporter` is only available HTTP :-/

juditnovak: (NOTE: The whole COS service alltogether is only over HTTP, meaning Prometheus, Kibana, etc. interfaces. So I think w'er not that bad with this :-) Even thought indeed it's a limitation.)

src/charm.py
juditnovak: Done.

phvalguima (approved):

tests/integration/helpers.py
```diff
@@ -139,6 +139,29 @@ def restart_unit(model_full_name: str, unit: str) -> None:
...
+    """Check if a given unit has 'kibana-exporter' service available and publishing."""
+    try:
+        # Normal IP address
+        socket.inet_aton(host)
+    except OSError:
+        socket.inet_pton(socket.AF_INET6, host)
+        host = f"[{host}]"
+
+    url = f"http://{host}:9684/metrics"
+    response = requests.get(url)
```
Mehdi-Bendriss, line 152: We need to handle http exceptions here - and return False if it fails

Mehdi-Bendriss (commented):
Thanks Judit - great work! I have 1 comment

tests/integration/helpers.py
juditnovak: Done.

Mehdi-Bendriss (approved):

Mehdi-Bendriss:
The issues raised from the HA network cut tests will be solved in a future PR this pulse

PR #99, opened 2024-08-01
[DPE-5024] Self hosted runners, Allure reports

juditnovak:


tests/integration/ha/test_scaling.py
```diff
@@ -33,6 +33,9 @@
 APP_AND_TLS = [APP_NAME, TLS_CERTIFICATES_APP_NAME]
 
 
+pytestmark = pytest.mark.runner(["self-hosted", "linux", "X64", "jammy", "large"])
+
+
```
Mehdi-Bendriss, line 38:
```suggestion
@pytest.mark.runner(["self-hosted", "linux", "X64", "jammy", "large"])
```

This should be repeated for every test in all integration test modules.

juditnovak: Indeed, fixed.

reneradoi (approved):

PR #103, opened 2024-08-13
Pin sync_docs.yaml version

a-velasco:
Pins `_sync_docs.yaml` to a release instead of `main`. This prevents the workflow from breaking when a new sync_docs is released to the public interface.


juditnovak (approved):
LGTM

PR #107, opened 2024-09-06
[DPE-5401] Keeping pipelines opensearch v214 until upgrade

juditnovak:
Keeping Opensearch version that's used on the pipelines on v2.14 (until [DPE-5400](https://warthogs.atlassian.net/browse/DPE-5400) is complete.

Also, in this PR: Fix for DP Worklows version inconsistency between CI and release pipelines (see https://github.com/canonical/opensearch-dashboards-operator/actions/runs/10613013476).

[DPE-5400]: https://warthogs.atlassian.net/browse/DPE-5400?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -17,14 +17,14 @@ jobs:
 
   build:
     name: Build charm
-    uses: canonical/data-platform-workflows/.github/workflows/build_charm.yaml@v17.0.1
+    uses: canonical/data-platform-workflows/.github/workflows/build_charm.yaml@v21.0.1
 
   release:
     name: Release charm
     needs:
       - ci-tests
       - build
-    uses: canonical/data-platform-workflows/.github/workflows/release_charm.yaml@v17.0.1
+    uses: canonical/data-platform-workflows/.github/workflows/release_charm.yaml@v21.0.1
     with:
       channel: 2/edge
       artifact-prefix: ${{ needs.build.outputs.artifact-prefix }}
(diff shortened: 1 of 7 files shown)


.github/workflows/release.yaml
```diff
@@ -17,14 +17,14 @@ jobs:
...
-    uses: canonical/data-platform-workflows/.github/workflows/build_charm.yaml@v16.7.0
+    uses: canonical/data-platform-workflows/.github/workflows/build_charm.yaml@v17.0.1
 
   release:
     name: Release charm
     needs:
       - ci-tests
       - build
-    uses: canonical/data-platform-workflows/.github/workflows/release_charm.yaml@v16.7.0
+    uses: canonical/data-platform-workflows/.github/workflows/release_charm.yaml@v17.0.1
```
phvalguima, line 27: Can we move to `v21.0.1` instead?

We are planning to do it for opensearch as well, here: https://github.com/canonical/opensearch-operator/pull/432

phvalguima (changes requested):

.github/workflows/release.yaml
juditnovak: We can move to any, but pls note that once Opensearch is on 2.16 OSD pipelines will break if this PR is not merged... While I'm on hoidays ;-) 

The above is surely fixing the broken Release pipelines we've got, which was the main objective.

I change it now, but I may ask you to make sure hte PR lands before releasing Opensearch 2.16.

Mehdi-Bendriss (approved):

phvalguima (approved):

PR #114, opened 2024-09-24
[DPE-4307] HA process interrupt tests

juditnovak:
This change unites 2 tasks:
 - [DPE-4307](https://warthogs.atlassian.net/browse/DPE-4307) **[OpenSearch Dashboards][VM] - Testing - HA process interrupt tests**
   - HA tests are covering _both_ OSD process intrrunpt and Opensearch Dashboards process interrupt
   - An issues was discovered and fixed:
     - OSD "hanigng" (simulated by sending `SIGSTOP` to the `node` process) was not handled by the charm. 
     - Healthcheck is extended, and a restart mechansim is added for the case when the application doesn't respond
 - [DPE-5516](https://warthogs.atlassian.net/browse/DPE-5516) **[OpenSearch Dashboards][VM][BUGFIX] Service unavailable until first `update-status`** 
   - **Reason**: 
     - I believe that the issue here was not to leave time for the service to establish, but triggering an `update-status` as soon as the process (i.e. snap level) was considered up. 
   - **Fix**:
     - A "grace period" was added, giving the application a chance to get back after a restart.
   - **Proof of functionality**:
     - See [earlier pipeline](https://github.com/canonical/opensearch-dashboards-operator/actions/runs/11039716508/job/30687023404) (not including the fix) with OSD showing regular `Service unavailable` status. 
     - Gone on current :green_circle: pipelines.

[DPE-4307]: https://warthogs.atlassian.net/browse/DPE-4307?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DPE-5516]: https://warthogs.atlassian.net/browse/DPE-5516?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/workload.py
```diff
@@ -109,9 +109,7 @@ def exec(self, command: list[str], working_dir: str | None = None) -> str:
     )
     def alive(self) -> bool:
         try:
-            return bool(self.dashboards.services[self.SNAP_APP_SERVICE]["active"]) and bool(
-                self.dashboards.services[self.SNAP_EXPORTER_SERVICE]["active"]
-            )
+            return bool(self.dashboards.services[self.SNAP_APP_SERVICE]["active"])
```
juditnovak, line 113: We had an ugly bug here. (@phvalguima is my witness :-) )

Reminder: OSD is a service that doesn't even come up (at all!) except in perfect conditions.
Meaning functional connection to Opensearch, etc. No, we do NOT even have a Status Page.
All we get is 
image

The Prometheus Exproter process is very similar. It can't come up as long as there is no functional OSD available. It's failing right after re-start otherwise.

Now this resulted in an ugly, flaky behavior whenever OSD _was_ considered `active` on a snap level -- yet the service was not running (for example: due to missing Opensearch connection). Meaning that the first part of the condition above was `True` -- while the 2nd part was `True/False` ... depending on the moment.

Screenshot from 2024-09-25 13-25-42

Debugging underlying data structures (representing snap service state):

Screenshot from 2024-09-25 13-25-54

The conclusion is this: we shouldn't check Prometheus Exporter state here. 

One proposal could be to do it in a separate check. (I can propose a solution even in parallel on a separte branch. If we want to address the matter, it would be much nicer in a separate task.). 
In my opinion, as long as we are not providing similar safety measures on Openserach (where monitoring is a WAY more crucial), it may be best to keep comlexity low and trus automatic snap restarts to cover Prometheus Exporter failures.

phvalguima: @juditnovak yes, I think we need to break the concepts: `alive()` here means the main service is up and running.

src/charm.py
```diff
@@ -151,9 +154,18 @@ def reconcile(self, event: EventBase) -> None:
         if getattr(event, "departing_unit", None) == self.unit:
             return
 
-        # 2. Restart on config change
+        # 2. Restart if the service is down or on config change
+        service_stopped = not self.workload.alive() and not self.unit.status == MaintenanceStatus(
```
Mehdi-Bendriss: In general we should avoid checking in the charm code the current status message. The check should rather be on the workload and **clear** the outdated status regardless of whether it was set or not.

Mehdi-Bendriss (commented):
Thanks Judit - I have a minor comment regarding the check on the status as part of the charm logic.

src/charm.py
phvalguima: Actually, I cannot see where that `MSG_ROLLING_RESTART` is really set...

src/charm.py
```diff
@@ -151,9 +154,18 @@ def reconcile(self, event: EventBase) -> None:
...
+            MSG_ROLLING_RESTART
+        )
+
+        # Evaluat unit health at this point (as it may trigger a restart)
+        unit_healthy, unit_msg = self.health_manager.unit_healthy()
+
         if (
-            self.config_manager.config_changed()
+            service_stopped
+            or (not unit_healthy and unit_msg == MSG_STATUS_HANGING)
```
phvalguima: We should remove this check

phvalguima (commented):
There is a potential loop here:

The reconcile method may end calling for a restart 

And, in the  `_restart`, [we end up calling reconcile again](https://github.com/canonical/opensearch-dashboards-operator/blob/2/edge/src/charm.py#L255-L262) before the actual restart.

A charm leader in a deployment with no lock assigned would end up in a continuous loop where the charm calls:
reconcile() > [detects unit is down](https://github.com/canonical/opensearch-dashboards-operator/pull/114/files#r1801382938) > calls `acquire_lock()` > which emits a [relation_changed](https://github.com/canonical/opensearch-dashboards-operator/blob/2/edge/lib/charms/rolling_ops/v0/rollingops.py#L399) > [assigns itself the lock](https://github.com/canonical/opensearch-dashboards-operator/blob/2/edge/lib/charms/rolling_ops/v0/rollingops.py#L348) > calls the [`_restart`](https://github.com/canonical/opensearch-dashboards-operator/blob/2/edge/src/charm.py#L255-L262) > re-calls the `reconcile` and restarts the loop


My recommendation is to restart the workload right away on `_restart` method.

src/charm.py
juditnovak: So... that's the thing.

This is basically our only (one of the few) notations of the charm currently being in the restart process. This message is set by the `rolling_ops` lib.

However, as you are highligting it -- it's indeed not right this way.
(It's kind of a hack to rely on a status set outside of the charm's code, etc.)
Let me try to think of a better way.

src/charm.py
phvalguima: This check is actually coming from self.health_manager.unit_healthy. We can continue with that.

juditnovak: Discussed offline: this check is NOT on the state of the charm, but a simple return value from a healthcheck. Therefore it's valid and should be left here.

Closing the thread with @phvalguima 's agreement.

src/charm.py
juditnovak: In fact, we don't need this check at all. Removed.

juditnovak:
@phvalguima You are right. Restart attempt on service down is removed from `reconcile()`. The check is not needed.

src/charm.py
```diff
@@ -254,22 +259,29 @@ def _start(self, event: EventBase) -> None:
...
+        # If the service is not yet functional, the status is set as
+        # 'Service unavailable' until the next 'update-status' hook execution
+        start_time = time.time()
+        unit_healthy, _ = self.health_manager.unit_healthy()
+        while not unit_healthy and time.time() - start_time < SERVICE_AVAILABLE_TIMEOUT:
+            time.sleep(5)
+            unit_healthy, _ = self.health_manager.unit_healthy()
+
         clear_status(self.unit, [MSG_STARTING, MSG_STARTING_SERVER])
-        self.on.update_status.emit()
```
phvalguima, line 279: Avoid a loop here, as described on: https://github.com/canonical/opensearch-dashboards-operator/pull/114#pullrequestreview-2369694313

There is no point in speeding up an `update-status`, it will happen anyways. In this case, it is better to leave some time between this `_restart` call and the next `update-status` for the system to settle.

phvalguima: Looking at this code again, may be worthy to structure it instead as follows:
```
        start_time = time.time()
        unit_healthy, _ = self.health_manager.unit_healthy()
        while not unit_healthy and time.time() - start_time < SERVICE_AVAILABLE_TIMEOUT:
            time.sleep(5)
            unit_healthy, _ = self.health_manager.unit_healthy()
        if unit_healthy:
            # We clean the status and finish
            clear_status(self.unit, [MSG_STARTING, MSG_STARTING_SERVER])
        else:
            # this call will potentially retrigger a restart
            self.on.update_status.emit()
```

phvalguima: FYI, the proposal above reintroduces the risk of continuous restart loop if the unit is failing because of some other problem...

phvalguima (dismissed):

Mehdi-Bendriss (dismissed):

Mehdi-Bendriss (approved):

PR #122, opened 2024-10-02
[DPE-5587] Fix private IP resolution to use juju spaces

### LINKED ISSUE #121: It always listens on 127.0.... and not on the node's IP when deploying to an external.
Hi, 
Opensearch-dashboard Listen on 127.0. and not on the fixed ip of the node opensearch-dashboard.
Despite the change to the /etc/hosts file.
Thank's. 

Capture d’écran du 2024-10-01 13-06-05

Capture d’écran du 2024-10-01 13-16-16

phvalguima:
Up to this point OSD was taking the local IP (using the Python `socket` library) for the application endpoint.

However we need to use the IP published by Juju instead (`Model.get_binding()`) -- as this is the one that can be contacted externally. 

The change has a impact on TLS certificates as well (as we are adjusting the IPs appearing in the certificate request accordingly).

Closes #121


phvalguima:
The result of this PR:
* From `src/core/models.py`, on `private_ip()`, we can see the following IP: `10.10.10.186` -> full PDB https://pastebin.ubuntu.com/p/j4mgtrV5Rr/
* This [is the /etc/hosts](https://pastebin.ubuntu.com/p/hPvRZYqNZq/)

Running the `socket` logic as we are doing currently on the charm indeed returns the wrong `127.0.1.1`:
```
root@juju-865fea-0:/var/lib/juju/agents/unit-opensearch-dashboards-0/charm# python3 
Python 3.10.12 (main, Sep 11 2024, 15:47:36) [GCC 11.4.0] on linux                  
Type "help", "copyright", "credits" or "license" for more information.              
>>> import socket                                                                   
>>> socket.gethostbyname(socket.gethostname())                                      
'127.0.1.1'                                                                         
>>>                                                                                 
```

src/core/models.py
```diff
@@ -171,19 +173,34 @@ def password_rotated(self) -> bool:
...
+        return hostnames[0]
 
     @property
     def private_ip(self) -> str:
         """The IP for the unit."""
-        return socket.gethostbyname(self.hostname)
+        # Either values are okay for us because we are interested on their name
+        # both objects have a .name property.
+        address = None
+        if binding := self.relation:
```
Mehdi-Bendriss: nit:
```suggestion
        if relation := self.relation:
```

src/core/models.py
```diff
@@ -171,19 +173,34 @@ def password_rotated(self) -> bool:
         return bool(self.relation_data.get("password-rotated", None))
 
     @property
-    def hostname(self) -> str:
+    def hostname(self) -> str | None:
         """The hostname for the unit."""
-        return socket.gethostname()
+        if not self.fqdn:
+            return None
+        return self.fqdn.split(".")[0]
```
Mehdi-Bendriss: why delete the `socket.hostname()`? I don't think this can be a problem

Please refer to the comment on `fqdn` below.

src/core/models.py
```diff
@@ -171,19 +173,34 @@ def password_rotated(self) -> bool:
...
         """The IP for the unit."""
-        return socket.gethostbyname(self.hostname)
+        # Either values are okay for us because we are interested on their name
+        # both objects have a .name property.
+        address = None
+        if binding := self.relation:
+            binding_obj = self.model.get_binding(binding.name)
+            address = (
+                str(binding_obj.network.bind_address)
+                if binding_obj and binding_obj.network
```
Mehdi-Bendriss: when can any of these 2 be ever `None`?

src/core/models.py
```diff
@@ -171,19 +173,34 @@ def password_rotated(self) -> bool:
...
+        return self.fqdn.split(".")[0]
 
     @property
-    def fqdn(self) -> str:
+    def fqdn(self) -> str | None:
         """The Fully Qualified Domain Name for the unit."""
-        return socket.getfqdn()
+        if not (hostnames := socket.gethostbyaddr(self.private_ip)):
+            return None
+        return hostnames[0]
```
Mehdi-Bendriss: Can you explain why we also stop relying on `socket.getfqdn()` ? 
For the following /etc/hosts:
```
127.0.0.1	localhost
255.255.255.255	broadcasthost
::1             localhost
```

Here are the outputs I'm getting:
```
> socket.getfqdn()

nebula.local
```

And for:
```
> socket.gethostname()

nebula.local
```

I think the problem is triggered by `socket.gethostbyname` (notice the `by`) when passing `127.0.0.1` as name. As it then reads from `/etc/hosts`

src/core/models.py
```diff
@@ -245,8 +262,8 @@ def sans(self) -> dict[str, list[str]]:
             return {}
 
         return {
-            "sans_ip": [self.private_ip],
-            "sans_dns": [self.hostname, self.fqdn],
+            "sans_ip": [ip for ip in [self.private_ip] if ip],
```
Mehdi-Bendriss:
```suggestion
            "sans_ip": self.private_ip,
```

src/core/models.py
```diff
@@ -245,8 +262,8 @@ def sans(self) -> dict[str, list[str]]:
...
+            "sans_dns": [dns for dns in [self.hostname, self.fqdn] if dns],
```
Mehdi-Bendriss:
```suggestion
            "sans_dns": [dns for dns in {self.hostname, self.fqdn} if dns],
```

src/literals.py
```diff
@@ -10,6 +10,7 @@
 CHARM_KEY = "opensearch-dashboards"
 
 PEER = "dashboard_peers"
+INGRESS_REL_NAME = "ingress"
```
Mehdi-Bendriss: to be removed.

Mehdi-Bendriss (commented):
Thanks pedro, I have some questions.

src/core/models.py
phvalguima: So, the flow here is the contrary right, you do *not* know about `127.0.1.1`. The question the `private_ip()` tries to answer is what is your IP. If you put:

> when passing 127.0.0.1 as name

then you already know your IP, why even bother asking to `private_ip`?

------------------------------------------------------------------------------------------------------------------

In reality, we can only discover our own IP either by two means:
1. get your hostname (readily present in `/etc/hostname`)  > resolve this name as a record
2. run `ip a` > get all the IPs > get the default route IP or the IP that has a route to your destination of interest

The (2) is most reliable; but (1) is the easiest by far.

If you want to ask what is the IP of your own hostname, then you will translate that hostname into address via DNS service in your box. That service is `systemd-resolved`. If you look into its [routing rules](https://www.freedesktop.org/software/systemd/man/latest/systemd-resolved.service.html#Protocols%20and%20Routing), you will see the first rule:

> Names for which synthetic records are generated (the local hostname, "localhost" and "localdomain", local gateway, as listed in the previous section) and addresses configured in /etc/hosts are never routed to the network and a reply is sent immediately.

So, if we have our own hostname defined in the /etc/hosts as `127.0.1.1`, this is the address you will get always.

PS.: you can see the resolving chain is configured in `resolved.conf` (not resolv.conf...) and [describes its default - which is the most used way in stock Ubuntu](https://www.freedesktop.org/software/systemd/man/latest/resolved.conf.html#Options)

-----------------------------------------------------------------------------------------------------

The most reliable way is method (2). That will really get you the IP. However, it can be tricky to figure out (1) the interface of interest; and (2) your target destination.

Therefore we have Juju. And there we can just grab the relevant IP.

src/core/models.py
phvalguima: @Mehdi-Bendriss it is because OSD has a way more strict linting tool. It essentially panics if it cannot resolve a given as a type or `None` (in this case, `binding_obj.network`). This was just a way around to tell the lint we will never have `None` on the subsequent line...

Personally, it is not really useful but I could not find a better way. Happy to hear any alternative

src/core/models.py
phvalguima: this is due to the lint

src/core/models.py
phvalguima: Interesting, I will commit this change and let's see.

src/core/models.py
phvalguima: Well, that is an interesting point. I deleted as in preparations for MAAS, we have to consider your server will have multiple records. So, MAAS registers the following:
* `eth0.myname.domain`  # resolves to the IP in eth0
* `eth1.myname.domain`  # resolves to the IP in eth1
* ...
* `myname.domain`: this record is specifically a repetition of one of the previous IPs

The `myname.domain`  will be the same IP as the default gateway of your host.

Anyways, it gets "nuanced" when moving to an env like maas, and fqdn may change depending on which interface you want to use.

phvalguima: So, asking for fqdn does not really matter in this context, as it depends always on the IP you are looking for.

src/core/models.py
```diff
@@ -171,31 +173,46 @@ def password_rotated(self) -> bool:
...
+        # Optional fallback, if no binding is found
+        return address or socket.gethostbyname(socket.gethostname())
 
     @property
     def public_ip(self) -> str:
-        result = subprocess.check_output(
-            [
-                "bash",
-                "-c",
-                "ip a | grep global | grep -v 'inet 10.' | cut -d' ' -f6 | cut -d'/' -f1",
```
phvalguima, line 194: This check will miss other [private IP ranges](https://datatracker.ietf.org/doc/html/rfc1918#section-3).
[Charm-helpers has a lot of facilities to check for IPs straight from interfaces, using netifaces](https://github.com/juju/charm-helpers/blob/1b2d4dc8f8effd79d782241a32a0485af1f01e73/charmhelpers/contrib/network/ip.py#L174-L212). 

However, I'd still stick with Juju-related information. It is very rare for a given workload to have its public IP attached to one interface. Hence, we need something that has a higher-level view of the entire deployment, like Juju.

skourta: Using juju can be misleading at times when a change of IP happens and it is yet to be picked by Juju.

src/core/models.py
```diff
@@ -171,31 +173,46 @@ def password_rotated(self) -> bool:
...
+        address = None
+        if relation := self.relation:
+            relation_obj = self.model.get_binding(relation.name)
+            address = (
+                str(relation_obj.network.bind_address)
+                if relation_obj and relation_obj.network
+                else None
+            )
+        # Optional fallback, if no binding is found
+        return address or socket.gethostbyname(socket.gethostname())
```
skourta: Woudn't it be better to use the property defined for this?

src/core/models.py
```diff
@@ -171,31 +173,46 @@ def password_rotated(self) -> bool:
...
 
     @property
     def private_ip(self) -> str:
         """The IP for the unit."""
-        return socket.gethostbyname(self.hostname)
+        # Either values are okay for us because we are interested on their name
+        # both objects have a .name property.
+        address = None
+        if relation := self.relation:
+            relation_obj = self.model.get_binding(relation.name)
```
skourta: I suggest we change the naming to reflect that this is a the binding of the relation

tests/integration/spaces/conftest.py
```diff
@@ -0,0 +1,125 @@
...
+        output = subprocess.run(
+            ["sudo", "lxc", "network", "show", name],
+            capture_output=True,
+            check=True,
+            encoding="utf-8",
+        ).stdout
+        logger.debug(f"LXD network status: {output}")
+
+        if not external:
+            subprocess.check_output(
```
skourta: Any reason why you switched from `run` to `check_output` I think the recommended option in `run` in the docs.

tests/integration/helpers.py
```diff
@@ -774,3 +795,20 @@ async def destroy_cluster(ops_test, app: str = OPENSEARCH_APP_NAME):
...
             # fails to restore the `update-status-hook-interval` value to it's former state.
             assert n_apps_after == n_apps_before - 1, "old cluster not destroyed successfully."
+
+
+async def for_machines(ops_test, machines, state="started"):
+    for attempt in Retrying(stop=stop_after_attempt(10), wait=wait_fixed(wait=60)):
+        with attempt:
+            for id in machines:
+                mach_status = json.loads(
+                    subprocess.check_output(
```
skourta: We should execute this outside the for loop

skourta (dismissed):
LGTM. Left some minor questions.

src/core/models.py
phvalguima: @skourta, in this case we should consider `netifaces` or any similar lib. However, this is the `public_ip` method. Rarely you will really have a public IP (e.g. 3.4.5.6) assigned to your local interfaces. These will generally be assigned to firewalls, load balancers, etc; any forwarder sitting in a DMZ.

src/core/models.py
```diff
@@ -171,31 +173,46 @@ def password_rotated(self) -> bool:
...
+            address = (
+                str(relation_obj.network.ingress_address)
+                if relation_obj and relation_obj.network
+                else None
+            )
+        # Optional fallback, if no binding is found
+        return address or socket.gethostbyname(socket.gethostname())
```
phvalguima: Likewise, to adopt `self.hostname` instead of `socket.gethostname()` here.

tests/integration/spaces/conftest.py
phvalguima: I don't really care about its output. That was the only reason tbh.

src/core/models.py
phvalguima: In this case where you have a security layer in a DMZ, then your "public_ip" will very likely be one among "10.0.0.0/8" or "172.16.0.0/12". Because the traffic will be arriving in the, e.g. firewall, and then forwarded to the datacenter network, which is all private IPs. 

Effectively, your "public IP" is a private IP from RFC 1918.

src/core/models.py
phvalguima: In discussions with @Mehdi-Bendriss, we've decided to park this particular change to `fqdn` and roll back to its original internals.

src/core/models.py
phvalguima: Actually, `binding_obj.network.bind_address` can be `None` if you have no binding set. The `binding_obj.network` issue is that `lint` cannot validate its type, hence it raises a failure as unknown type or None. That is why I am adding this second part of the check.

src/core/models.py
phvalguima: @Mehdi-Bendriss did I answer your question here?

src/core/models.py
phvalguima: Returned to `"sans_ip": [self.private_ip],`

tests/integration/helpers.py
```diff
@@ -31,6 +31,23 @@
...
+        - [ 'sysctl', '-w', 'fs.file-max=1048576' ]
+        - [ 'sysctl', '-w', 'vm.swappiness=0' ]
+        - [ 'sysctl', '-w', 'net.ipv4.tcp_retries2=5' ]
+    """,
+}
+
+TLS_CERTIFICATES_APP_NAME = "self-signed-certificates"
+COS_AGENT_APP_NAME = "grafana-agent"
+COS_AGENT_RELATION_NAME = "cos-agent"
+DB_CLIENT_APP_NAME = "application"
```
juditnovak, line 51: This block doesn't seem to be used overall (OPENSEARCH_CONFIG for instance).

Could we just add as much as required here? Thx much :-)

src/core/models.py
```diff
@@ -61,7 +61,7 @@ class OpensearchServer(StateBase):
 
     def __init__(
         self,
-        relation: Relation | None,
+        relation: Relation,
```
juditnovak: The Opensearch connection may disappear at any point in the charm's life.

We shouldn't get exceptions if passing `None` here on such occasions. The charm code is structured in a way that such instantiation of `OpensearchServer()` is accepted.

src/core/cluster.py
```diff
@@ -12,7 +12,7 @@
     OpenSearchRequiresData,
 )
 from ops.framework import Framework, Object
-from ops.model import Relation, Unit
+from ops.model import Binding, Relation, Unit
```
juditnovak: Is `Binding` used?

src/core/models.py
```diff
@@ -142,13 +142,15 @@ class ODServer(StateBase):
 
     def __init__(
         self,
+        model: Model,
```
juditnovak: Please, let's try to avoid doing this if possible...

Let me highlight regarding the structure of "new-style" charms.
There has been a strong Separation of Concerns applied, typically to avoid "almighty" components such as `model` or `charm` wired in spaces where they aren't neeed. 
The `src/core/models` module implements the representation of Charm Components (such as `Unit`, `App`, etc.) based on Relation Data or very basic additional, independent actions (such as `socket` calls).

To my understanding, binding information is outside of the scope of the `ODServer` object, as it only depends on the `ClusterState` object (implementing the `peer_relation` property). Thus I rather suggest a `peer_binding` in `src/core/cluster` (see example on [other "new-style" charm](https://github.com/canonical/kafka-operator/blob/7cfebeee4066b0258dde0560a70b9c68c7ebca7a/src/core/cluster.py#L296-L303
) ).

Now a next quesetion is if we want to use the new `osd_charm.state.bind_address` in all places where we've been using `private_ip` before? If yes, we may want to apply corresponding changes on `events/tls` and `managers/tls`. Otherwise changing `managers/config` is enough.

src/core/cluster.py
phvalguima: Not anymore, removed

skourta (approved):

src/core/models.py
phvalguima: This is not the case anymore.

Mehdi-Bendriss (approved):

PR #125, opened 2024-10-19
[DPE-5704] Update COS to use jammy series

phvalguima:
Current COS defaults to 20.04, which is not supported anymore by juju 3.6.

This PR updates the COS deployment to use 22.04 series instead.


juditnovak (approved):
All looks good, but may it be possible to also upgrade outdated libs?
Just for nicer pipelines statistics...

Seems like a new TLS lib version. Thx very-very much...

skourta (approved):

Mehdi-Bendriss (approved):

phvalguima:
@juditnovak I will update the lib on a follow-up

Issue #35, opened 2024-05-07
Introduce YAML parser for dealing with yaml config operations


Mehdi-Bendriss:
Currently YAML related operations (read / write) is handled as plain text operations. 
We should use proper yaml parsing when dealing with yaml documents.

juditnovak:
This was a design decision that have been discussed multiple times.

I see no need for YAML operations in the current layout (and I'm not the only one with that view in the Data Platform Team).

Aslo, may I ask why a know design decision would you label as a bug...? 

Is there a missing functionality that can not be fullfilled with the current method?

In other words: what is the purpose of this open issue? What is the problem we are trying to resolve?

(Part of the reason why I'm asking is because we have a few burning issues around the Dashboards...)

PR #131, opened 2024-11-08
[DPE-5909][DPE-4308] Support secure_authentication toggle:

### LINKED ISSUE #129: Modern browsers fail to login with HTTP-only dashboards
Modern browsers, in my case, both:
* Mozilla Firefox 131.0.2
* Google Chrome 128.0.6613.119 

Result in opensearch-dashboards never leaving the login page when running with HTTP only.

There is nothing relevant in the logs, besides:
```
{"type":"response","@timestamp":"2024-11-05T14:42:30Z","tags":[],"pid":5161,"method":"post","statusCode":401,"req":{"url":"/api/ism/accountInfo","method":"post","headers":{"host":"10.235.113.227:5601","user-agent":"Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:131.0
) Gecko/20100101 Firefox/131.0","accept":"*/*","accept-language":"en-US,en;q=0.5","accept-encoding":"gzip, deflate","referer":"http://10.235.113.227:5601/app/login?","content-type":"application/json","osd-version":"2.17.0","osd-xsrf":"osd-fetch","content-length":"82","origin":"http://10.235.113.227:5601","dnt":"1","connection":"keep-alive","priority":"u=4"},"remoteAddress":"10.235.113.1","userAgent":"Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:131.0) Gecko/20100101 Firefox/131.0","referer":"http://10.235.113.227:5601/app/
login?"},"res":{"statusCode":401,"responseTime":3,"contentLength":9},"message":"POST /api/ism/accountInfo 401 3ms - 9.0B"}
```

Looking closer with Firefox's dev tools, I can see a console error thou:
```
Cookie “security_authentication” has been rejected because a non-HTTPS cookie can’t be set as “secure”.
```

So, I believe we are missing to submit a token due to modern browsers' security restrictions and that causes the server to issue a 401 Unauthorized.

My recommendation is to deprecate HTTP and support only HTTPS moving forward.

phvalguima:
The current charm enforces secure_authentication cookie to True, which means we cannot effectively use opensearch-dashboards without TLS. This change extends the current charm to instead toggle the secure_authentication and other TLS-related options accordingly.

This change also reformats config_manager.py to load and manage configuration as a dict instead of a list.

Closes #129, #35

### DIFF
--- a/src/managers/config.py
+++ b/src/managers/config.py
@@ -4,8 +4,9 @@
 
 """Manager for for handling configuration building + writing."""
 import logging
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Any
 
+import yaml
 from ops.model import ConfigData
 
 if TYPE_CHECKING:
@@ -17,22 +18,26 @@ from core.workload import WorkloadBase
 logger = logging.getLogger(__name__)
 
 
-DEFAULT_PROPERTIES = """
-opensearch.ssl.verificationMode: full
-opensearch.requestHeadersWhitelist: [authorization, securitytenant]
-opensearch_security.multitenancy.enabled: true
-opensearch_security.multitenancy.tenants.preferred: [Private, Global]
-opensearch_security.readonly_mode.roles: [kibana_read_only]
-opensearch_security.cookie.secure: true
-"""
+DEFAULT_PROPERTIES = {
+    "opensearch.requestHeadersWhitelist": ["authorization", "securitytenant"],
+    "opensearch_security.multitenancy.enabled": True,
+    "opensearch_security.multitenancy.tenants.preferred": ["Private", "Global"],
+    "opensearch_security.readonly_mode.roles": ["kibana_read_only"],
+    "server.ssl.enabled": False,
+    "opensearch_security.cookie.secure": False,
+    "server.ssl.supportedProtocols": ["TLSv1.2"],  # In tandem with current OpenSearch
+}
 
-TLS_PROPERTIES = """
-server.ssl.enabled: true
-"""
+# Overrides the DEFAULT_PROPERTIES if we have TLS enabled
+TLS_PROPERTIES = {
+    "server.ssl.enabled": True,
+    "opensearch.ssl.verificationMode": "full",
+    "opensearch_security.cookie.secure": True,
+}
 
-LOG_PROPERTIES = """
-logging.verbose: true
-"""
+LOG_PROPERTIES = {
+    "logging.verbose": True,
+}
 
 
 class ConfigManager:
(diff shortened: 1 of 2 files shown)


src/managers/config.py
```diff
@@ -17,22 +18,26 @@
...
-opensearch_security.cookie.secure: true
-"""
+DEFAULT_PROPERTIES = {
+    "opensearch.requestHeadersWhitelist": ["authorization", "securitytenant"],
+    "opensearch_security.multitenancy.enabled": True,
+    "opensearch_security.multitenancy.tenants.preferred": ["Private", "Global"],
+    "opensearch_security.readonly_mode.roles": ["kibana_read_only"],
+    "server.ssl.enabled": False,
+    "opensearch_security.cookie.secure": False,
+    "server.ssl.supportedProtocols": ["TLSv1.2"],  # In tandem with current OpenSearch
```
Mehdi-Bendriss: In OpenSearch we are unfortunately stuck with TLSv1.2 because of the CA rotation topic. But in the dashboards we are not, I believe we can delete this, which would default to `['TLSv1.1', 'TLSv1.2', and 'TLSv1.3']`

src/managers/config.py
```diff
@@ -139,61 +140,26 @@ def current_properties(self) -> list[str]:
...
-        config_properties = self.static_properties
-
-        properties_changed = set(server_properties) ^ set(config_properties)
-
-        if not properties_changed:
+        if self.load_dashboard_properties() == self.dashboard_properties:
             return False
-
-        if properties_changed:
-            logger.info(
```
phvalguima, line 190: Properties have sensitive information, we should not log it. We do not do these diffs for OpenSearch either.

src/managers/config.py
```diff
@@ -139,61 +140,26 @@ def current_properties(self) -> list[str]:
...
     def set_dashboard_properties(self) -> None:
         """Writes built config file."""
         self.workload.write(
-            content="\n".join(self.dashboard_properties),
+            content=yaml.dump(self.dashboard_properties),
             path=self.workload.paths.properties,
         )
 
-    @staticmethod
-    def build_static_properties(properties: list[str]) -> list[str]:
```
phvalguima, line 164: not used

src/managers/config.py
```diff
@@ -139,61 +140,26 @@ def current_properties(self) -> list[str]:
         """The current configuration properties set to zoo.cfg."""
         return self.workload.read(self.workload.paths.properties)
 
-    @property
-    def current_env(self) -> list[str]:
-        """The current /etc/environment variables."""
-        return self.workload.read(path="/etc/environment")
-
-    @property
-    def static_properties(self) -> list[str]:
```
phvalguima, line 148: not used

src/managers/config.py
```diff
@@ -139,61 +140,26 @@ def current_properties(self) -> list[str]:
         """The current configuration properties set to zoo.cfg."""
         return self.workload.read(self.workload.paths.properties)
 
-    @property
-    def current_env(self) -> list[str]:
```
phvalguima, line 143: not used

Mehdi-Bendriss (commented):
Thanks Pedro, looks good! I have 1 minor comment on the TLS version

Mehdi-Bendriss (approved):

PR #136, opened 2024-11-26
Remove dependency breaking ccc-hub

phvalguima:
The `poetry-plugin-export` seems to be causing issues with newer ccc-hub. This PR removes this dependency and updates the poetry lock file.

This PR disables the cache. We will need a 2nd round where we re-enable it.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -72,7 +72,7 @@ jobs:
     uses: canonical/data-platform-workflows/.github/workflows/build_charm.yaml@v22.0.0
     with:
       path-to-charm-directory: ${{ matrix.path }}
-      cache: true
+      cache: false
 
   integration-test:
     strategy:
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yaml
```diff
@@ -72,7 +72,7 @@ jobs:
     uses: canonical/data-platform-workflows/.github/workflows/build_charm.yaml@v22.0.0
     with:
       path-to-charm-directory: ${{ matrix.path }}
-      cache: true
+      cache: false
```
dragomirp, line 75: ~~Should also set `artifact-prefix` to somthing different from `packed-charm-cache`~~ Prefix is as it is due to there being a build matrix for the regular charm and the test charm.

dragomirp (approved):

Mehdi-Bendriss (approved):

PR #142, opened 2024-12-10
Add promtool check and test in the CI

gabrielcocenza:
This PR adds the check and test of Prometheus rules. Because the exporter is going to change and also the metrics/rules, the only unit test covered is for if the scrape is up.

When the new exporter is used, then the new rules and unit tests will be implemented.

## How to test this PR

1. Install the prometheus snap:
```shell
sudo snap install prometheus
```

2. Run to check if the rules file is valid:
```shell
promtool check rules src/alert_rules/prometheus/*.yaml
```

3. Run the unit tests:
```
promtool test rules tests/unit/test_alert_rules/*.yaml
```

### DIFF
--- a/.github/workflows/test_prometheus_rules.yaml
+++ b/.github/workflows/test_prometheus_rules.yaml
@@ -3,12 +3,6 @@ name: Test prometheus rules
 on:
   workflow_call:
   workflow_dispatch:
-  pull_request:
-    types: [opened, synchronize, reopened]
-    branches: [main]
-    paths-ignore:
-      - "**.md"
-      - "**.rst"
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.head_ref || github.ref }}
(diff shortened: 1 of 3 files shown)


zmraul (changes requested):
Very interesting, didn't know this was a thing. I would prefer this check to be part of general CI runs, not a workflow on its own.

So probably remove `on: pull_request` and just call this workflow on `ci.yaml`

phvalguima (dismissed):

.github/workflows/test_prometheus_rules.yaml
```diff
@@ -0,0 +1,28 @@
+name: Test prometheus rules
```
zmraul, line 1: This needs to be called on `ci.yaml` no?:
```yaml
# ci.yaml
jobs:
  unit-tests:
  . . .
  alert-unit-tests:
    name: Alert tests
    uses: ./.github/workflows/test_prometheus_rules.yaml
```

gabrielcocenza: Sorry, for some reason I forgot to add this change.

zmraul (approved):

phvalguima (approved):

PR #159, opened 2025-01-22
Add new Prometheus Rules using the python exporter

gabrielcocenza:
Critical alerts for:
- dashboards in red state
- dashboards plugins in red state
- high response latency
- no metrics generation from the exporter

Warning alerts for:
- dashboards in yellow state
- dashboards plugins in yellow state

- Added unit tests for the new rules

I've manually tested and the metrics are working as expected and stopping the OpenSearch dashboars service triggers alert as expected:

image

It was also necessary to change the dashboards to use the new metrics:
image


Mehdi-Bendriss (dismissed):
Thanks Gabriel, great work!

Mehdi-Bendriss (approved):

PR #164, opened 2025-02-06
[DPE-6536] Improve terraform charm module

Mehdi-Bendriss:
This PR implements [DPE-6536](https://warthogs.atlassian.net/browse/DPE-6536). Namely, this PR: 
- improves the existing terraform charm module
- removes dependency on opensearch 
- fixes the `model` as per the [TF spec](https://docs.google.com/document/d/1EG71A2pJ244PQRaGVzGj7Mx2B_bzE4U_OSqx4eeVI1E/edit?tab=t.0) 
- fixes the output variables as per the [TF spec](https://docs.google.com/document/d/1EG71A2pJ244PQRaGVzGj7Mx2B_bzE4U_OSqx4eeVI1E/edit?tab=t.0) 
- add conditional setup of TLS

[DPE-6536]: https://warthogs.atlassian.net/browse/DPE-6536?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/workflows/ci.yaml
```diff
@@ -128,8 +123,8 @@ jobs:
           juju add-model test
       - name: Terraform deploy
         run: |
-          pushd ./terraform/tests/
-          TF_VAR_model_name="test" terraform apply -target null_resource.simple_deployment_juju_wait_deployment -auto-approve
+          pushd ./terraform
+          terraform apply -var='model=test' -auto-approve
```
phvalguima, line 123: This is not going to do much besides start a deployment and immediately delete it later. The `juju-wait` was there to ensure the deployment comes up.

terraform/main.tf
```diff
@@ -33,3 +33,35 @@ resource "juju_application" "opensearch-dashboards" {
     }
   }
 }
+
+# Deploy the self-signed-certificates operator if tls enabled
+resource "juju_application" "self-signed-certificates" {
+  for_each = var.tls ? { "deployed" = true } : {}
```
phvalguima, line 39: nice

terraform/outputs.tf
```diff
@@ -6,21 +6,18 @@ output "app_name" {
...
-
-output "opensearch_client_endpoint" {
-  description = "Name of the endpoint opensearch-client endpoint."
-  value       = "opensearch-client"
+# integration endpoints
+output "requires" {
+  description = "Map of all \"requires\" endpoints"
+  value = {
+    opensearch_client = "opensearch-client"
+    certificates      = "certificates"
```
phvalguima, line 14: There is a risk here someone creates a 2nd relation for certificates & sets `var.tls = true` at the same time. This can happen b/c "requires.certificates" is just a string that anyone else can use.

I'd suggest we have it in `provides` instead, then we can "consume" the self-signed-certificates endpoint instead of providing an endpoint to connect with OSD.

terraform/variables.tf
```diff
@@ -59,3 +59,10 @@ variable "endpoint_bindings" {
   type        = map(string)
   default     = {}
 }
+
+# additional variables
+variable "tls" {
+  description = "Whether TLS should be enabled"
+  type        = bool
+  default     = false
+}
```
MiaAltieri, line 68: newb opensearch question: isn't it a requirement that opensearch have TLS enabled? Is that also true for its dashboards?

terraform/variables.tf
```diff
@@ -59,3 +59,10 @@ variable "endpoint_bindings" {
   type        = map(string)
   default     = {}
 }
+
```
MiaAltieri, line 62: QQ - why have a TLS variable and the tls integration? Isn't that counterintuitive?

MiaAltieri (commented):
Looks good, minor questions only.

[Optional: a sanity check test]

terraform/outputs.tf
Mehdi-Bendriss: According to the [TF Spec](https://docs.google.com/document/d/1EG71A2pJ244PQRaGVzGj7Mx2B_bzE4U_OSqx4eeVI1E/edit?tab=t.0) this output variable should in-fact reflect the provides/requires endpoints as defined in the `metadata.yaml`

terraform/variables.tf
Mehdi-Bendriss: OpenSearch yes, opensearch-dashboards (the current) however, no.

terraform/variables.tf
Mehdi-Bendriss: The TLS resource + integration are optional and conditioned to the `tls` variable to be `true`, meaning only if the tls flag is set to true would the self-signed-certs app be deployed and integrated to the dashboards
e.g:
```
resource "juju_application" "self-signed-certificates" {
  for_each = var.tls ? { "deployed" = true } : {}

resource "juju_integration" "tls-opensearch_dashboards_integration" {
  for_each = var.tls ? { "deployed" = true } : {}
```

.github/workflows/ci.yaml
Mehdi-Bendriss: Good catch, I added the native `juju wait-for`

MiaAltieri (approved):

phvalguima (approved):

PR #173, opened 2025-03-27
[DPE-6858][TF] Add expose support

phvalguima:
This PR adds `expose` as a variable and subsequent dynamic block. If the expose is set to true, then the dynamic block renders `expose {}`, which effectively exposes the application to any external access.

Terraform plan with `expose=true`: https://pastebin.ubuntu.com/p/M7RvwmSK22/
Terraform plan without expose (hence defaulting to `false`): https://pastebin.ubuntu.com/p/NTPGzcMPP7/

### DIFF
--- a/terraform/main.tf
+++ b/terraform/main.tf
@@ -15,6 +15,11 @@ resource "juju_application" "opensearch-dashboards" {
   units       = var.units
   constraints = var.constraints
 
+  dynamic "expose" {
+    for_each = var.expose ? [1] : []
+    content {}
+  }
+
   # TODO: uncomment once final fixes have been added for:
   # Error: juju/terraform-provider-juju#443, juju/terraform-provider-juju#182
   # placement = join(",", var.machines)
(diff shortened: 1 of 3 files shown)


terraform/variables.tf
```diff
@@ -65,6 +65,12 @@ variable "endpoint_bindings" {
   default     = {}
 }
 
+variable "expose" {
+  description = "Expose the application"
```
Mehdi-Bendriss, line 69:
```suggestion
  description = "Expose the application for external access."
```

Mehdi-Bendriss (approved):
Thanks Pedro

terraform/main.tf
```diff
@@ -15,6 +15,11 @@ resource "juju_application" "opensearch-dashboards" {
   units       = var.units
   constraints = var.constraints
 
+  dynamic "expose" {
+    for_each = var.expose ? [1] : []
+    content {}
+  }
```
marcoppenheimer, line 21: **question:** What are we trying to achieve here? K8s exposure? Or just plain Juju expose? Where do we need to `juju expose`?

phvalguima: Plain `juju expose`

marcoppenheimer (approved):

PR #180, opened 2025-04-24
[DPE-6091] Add oauth integration

Zvirovyi:
# Overview

This PR adds OAuth integration to allow authorization in the OpenSearch Dashboards through SSO using the clients from the [Canonical Identity Platform](https://charmhub.io/identity-platform).

You can find example of use in the integration test.

# Key notes

* TLS should be enabled to avoid security errors from the `hydra` (`identity-platform`), and both of them should use the same certificates (correspondingly from the single `self-signed-certificates`);
* this PR depends on the [same PR for the opensearch](https://github.com/canonical/opensearch-operator/pull/612). Until it's merged, oauth integration test will fail. In order to run it locally you will need to use locally builded `opensearch` charm.


tests/integration/test_oauth.py
```diff
@@ -0,0 +1,267 @@
...
+
+    action_output = await get_redirect_uri_action.wait()
+    assert "redirect-uri" in action_output.results
+
+    ext_idp_service.update_redirect_uri(action_output.results["redirect-uri"])
+
+    global opensearch_dashboards_ip
+    opensearch_dashboards_ip = await get_address(
+        ops_test, ops_test.model.applications[APP_NAME].units[0].name
+    )
```
nsklikas: This shouldn't be needed, the test lib should be taking care of configuring the integrator charm. Was this copy/pasted or did you face any issues?

src/events/oauth.py
```diff
@@ -0,0 +1,51 @@
...
+class OAuthHandler(Object):
+    """Handler for managing oauth relations."""
+
+    def __init__(self, charm) -> None:
+        super().__init__(charm, "oauth")
+        self.charm: "OpensearchDasboardsCharm" = charm
+
+        client_config = ClientConfig(
+            audience=["opensearch"],
+            redirect_uri=f"{self.charm.state.url}/auth/openid/login",
```
nsklikas: Not sure if this can happen, BUT if `self.charm.state.url` changes then you will need to call `self.oauth.update_client_config` to update the url in the databag. This is not handled by the lib, not sure if it is something that we should add. Feel free to open an feature request in [hydra](https://github.com/canonical/hydra-operator/issues) if you think we should have it.

nsklikas (approved):
Skimmed through the PR and everything looks great, just have a couple of comments

tests/integration/test_oauth.py
```diff
@@ -0,0 +1,267 @@
...
+@pytest.mark.runner(["self-hosted", "linux", "X64", "jammy", "large"])
+@pytest.mark.group(1)
+@pytest.mark.abort_on_fail
+@pytest.mark.skip_if_deployed
+async def test_deploy_identity_bundle(
+    ops_test: OpsTest, ops_test_microk8s: OpsTest, ext_idp_service: ExternalIdpService
+):
+    await deploy_identity_bundle(
+        ops_test=ops_test_microk8s,
+        bundle_channel="latest/edge",
```
nsklikas, line 76: I would suggest to use `latest/stable`, in order to avoid running into bugs.

src/events/oauth.py
Zvirovyi: Yes, `redirect_uri` can change due to TLS changes (http / https switch).
So, are you telling that even when we add `update_client_config`, hydra will not be able to process it properly?

tests/integration/test_oauth.py
Zvirovyi: There is no stable release for that bundle:
```
$ juju info identity-platform
name: identity-platform
publisher: Identity Charmers
bundle-id: aG9JtaPHvtvE8ZNtN59Sg9MZWXqjgr5I
channels: |
  latest/stable:     –
  latest/candidate:  –
  latest/beta:       –
  latest/edge:       39  2025-01-14  (39)  6kB  @
  istio/stable:      –
  istio/candidate:   –
  istio/beta:        –
  istio/edge:        38  2025-01-12  (38)  6kB  @
  0.3/stable:        –
  0.3/candidate:     –
  0.3/beta:          –
  0.3/edge:          32  2024-09-20  (32)  6kB  @
  0.2/stable:        –
  0.2/candidate:     –
  0.2/beta:          –
  0.2/edge:          25  2024-05-09  (25)  5kB  @
  0.1/stable:        –
  0.1/candidate:     –
  0.1/beta:          –
  0.1/edge:          17  2024-04-25  (17)  5kB  @
```

tests/integration/test_oauth.py
Zvirovyi: Re-tested and indeed - it's not needed. I added this to workaround issue "wrong code" before I implemented proper `opensearch_security.openid.base_redirect_url` config parameter and a fix for the missing client secret.

src/events/oauth.py
Zvirovyi: I did further testing and it turns out to be that this charm cannot automatically switch between http / https modes. Therefore, in my opinion, both `update_client_config` and http / https switching are out of scope of this PR and should be implemented as separate issue.

Mehdi-Bendriss: @Zvirovyi could you add that call and comment it and add a comment about the context, we will address the automated switch from tls to non-tls in the coming weeks

src/events/oauth.py
```diff
@@ -0,0 +1,60 @@
...
+            self.charm.on[OAUTH_REL_NAME].relation_changed, self._on_oauth_relation_changed
+        )
+        self.framework.observe(
+            self.charm.on[OAUTH_REL_NAME].relation_broken, self._on_oauth_relation_changed
+        )
+
+    def _on_oauth_relation_changed(self, event: EventBase) -> None:
+        """Handler for `_on_oauth_relation_changed` event."""
+        if not self.charm.state.servers:
+            return
```
Mehdi-Bendriss, line 40: we should perhaps defer before returning, right?

tests/integration/test_oauth.py
Mehdi-Bendriss, line 1: Missing docstrings and inline comments to document the flow

Mehdi-Bendriss (commented):
Thank you Vladyslav, I manually tested the flow and it works perfectly. Left small comments.
We should be good to go once the CI passes

tests/integration/test_oauth.py
Zvirovyi: Done

src/events/oauth.py
Zvirovyi: Done

src/events/oauth.py
Zvirovyi: Done. Added `TODO` comment about it.

src/core/models.py
```diff
@@ -239,3 +241,65 @@ def sans(self) -> dict[str, list[str]]:
             "sans_ip": [self.private_ip, self.public_ip],
             "sans_dns": [dns for dns in {self.hostname, self.fqdn} if dns],
         }
+
+    @property
+    def oauth_client_secret(self) -> str:
+        """Client secret of the Oauth relation."""
+        return self.relation_data.get("oauth-client-secret", "")
```
zmraul: **todo**: Small change, `oauth-client-secret` needs adding to `PEER_APP_SECRETS` in `literals.py` so it's stored as a secret by the charm, not as plaintext on the databag. Maybe this should go into app level databag, since it's the same for the whole app. So `ODCluster` class.

zmraul (changes requested):
Very good thanks! 

Only small change that needs addressing is storing the oauth secret as a secret instead of doing so in the databag.

src/core/models.py
Zvirovyi: Done

Mehdi-Bendriss (dismissed):

zmraul (dismissed):

zmraul:
PR looks good, but the tests are failing at the setup phase, not related to the test itself. Once that is fixed we should be good to merge.

Mehdi-Bendriss (approved):

PR #204, opened 2025-07-23
Add refresh_versions.toml

smethnani:
The tag_charm_edge.yaml workflow expects a file called `refresh_versions.toml`. This PR adds that file


smethnani:
We should in fact omit the `tag` job according to the workflow [docs](https://github.com/canonical/data-platform-workflows/blob/main/.github/workflows/release_charm_edge.md). Thank you @carlcsaposs-canonical!

PR #213, opened 2025-09-01
[DPE-7927] - Disable TLS on relation broken

skourta:
This pull request updates the TLS certificate handling logic to remove the certificates and disable TLS when the TLS relation is broken. It also adjusts the corresponding unit tests to reflect the new behavior.

TLS certificate removal:

* Modified the `_on_certs_relation_broken` method in `src/events/tls.py` to always remove certificates when the relation is broken, regardless of certificate validity.

Unit test updates:

* Updated `test_certificates_broken` in `tests/unit/test_tls.py` to assert that all certificate-related fields (`certificate`, `ca`, `csr`, `tls`) are cleared when the TLS relation is removed.
* Removed an unused import of `TLSEvents` from `tests/unit/test_tls.py`.

### DIFF
--- a/src/events/tls.py
+++ b/src/events/tls.py
@@ -144,8 +144,7 @@ class TLSEvents(Object):
     def _on_certs_relation_broken(self, _) -> None:
         """Handler for `certificates_relation_broken` event."""
         # In case we have valid certificates, we keep them for smooth service function
-        if self.charm.state.unit_server.tls and not self.charm.tls_manager.certificate_valid():
-            self._remove_certificates()
+        self._remove_certificates()
 
     def _set_tls_private_key(self, event: ActionEvent) -> None:
         """Handler for `set-tls-privat-key` event when user manually specifies private-keys for a unit."""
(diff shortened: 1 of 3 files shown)


src/events/tls.py
```diff
@@ -144,8 +144,7 @@ def _on_config_changed(self, event: EventBase):
     def _on_certs_relation_broken(self, _) -> None:
         """Handler for `certificates_relation_broken` event."""
         # In case we have valid certificates, we keep them for smooth service function
-        if self.charm.state.unit_server.tls and not self.charm.tls_manager.certificate_valid():
-            self._remove_certificates()
+        self._remove_certificates()
```
Mehdi-Bendriss, line 147: Does this not require a restart to re-triggered? through the reconcile method or similar.

Mehdi-Bendriss (commented):
Thanks Smail. I have a question.

src/events/tls.py
skourta: It will make a restart through the reconcile method that will be called on `secret_changed` when the CSR, certificate and CA are removed.

smethnani (approved):
Thanks @skourta! Tested manually and removing the tls relation triggered a restart and I was able to access dashboards via http.

Mehdi-Bendriss (approved):

PR #214, opened 2025-09-08
[DPE-7774]: Raise if error on install

### LINKED ISSUE #202: Charm does not enter error state nor stop execution flow when it cannot install the workload.
# How to reproduce

- Set OPENSEARCH_DASHBOARDS_SNAP_REVISION to a non existing revision in literals.py.
- Pack the charm.
- Deploy it.

During deployment the charm will log the error during the installation:
```
2025-07-17T11:03:50.4290610Z unit-opensearch-dashboards-1: 10:59:31 ERROR unit.opensearch-dashboards/1.juju-log Snap: 'opensearch-dashboards'; command ['snap', 'install', 'opensearch-dashboards', '--channel="xxx"', '--revision="yyy"'] failed with output = ''
2025-07-17T11:03:50.4291063Z unit-opensearch-dashboards-1: 10:59:32 ERROR unit.opensearch-dashboards/1.juju-log dashboard_peers:0: Snap: 'opensearch-dashboards'; command ['snap', 'install', 'opensearch-dashboards', '--channel="xxx"', '--revision="yyy"'] failed with output = ''
```

And will fail on the next action requiring the snap to be deployed, currently:
```
2025-07-17T11:03:50.4295952Z   File "/var/lib/juju/agents/unit-opensearch-dashboards-1/charm/src/workload.py", line 97, in write
2025-07-17T11:03:50.4296092Z     shutil.chown(os.path.dirname(path), user="snap_daemon", group="root")
2025-07-17T11:03:50.4296200Z   File "/usr/lib/python3.10/shutil.py", line 1374, in chown
2025-07-17T11:03:50.4296304Z     raise LookupError("no such user: {!r}".format(user))
2025-07-17T11:03:50.4296393Z LookupError: no such user: 'snap_daemon'
```

# Expected behavior

The charm should stop the execution flow as soon as the install is failing.

skourta:
This pull request improves error handling during the installation process of Opensearch Dashboards by introducing a custom exception for installation failures and updating related unit tests to ensure correct behavior. 

fixes #202 

**Error Handling Improvements:**
* Added a new `OSDInstallError` exception in `src/exceptions.py` to specifically handle Opensearch Dashboards installation issues.
* Updated `src/charm.py` to raise `OSDInstallError` when the workload installation fails. [[1]](diffhunk://#diff-b9ed39bbc9c0387bd3e07da31d13373745534a1cd723d3e292c73496b12e307cR22) [[2]](diffhunk://#diff-b9ed39bbc9c0387bd3e07da31d13373745534a1cd723d3e292c73496b12e307cR138-R140)

**Testing Updates:**
* Updated the `test_install_blocks_snap_install_failure` test to expect the new exception when installation fails, ensuring the charm properly blocks and reports errors.
* Corrected the `test_install_sets_ip_hostname_fqdn` test to patch the install method with a successful return value, aligning with the intended test scenario.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -19,6 +19,7 @@ from events.oauth import OAuthHandler
 from events.requirer import RequirerEvents
 from events.tls import TLSEvents
 from events.upgrade import ODUpgradeEvents, OpensearchDashboardsDependencyModel
+from exceptions import OSDInstallError
 from helpers import (
     clear_global_status,
     clear_status,
@@ -134,6 +135,9 @@ class OpensearchDashboardsCharm(CharmBase):
         install = self.workload.install()
         if not install:
             self.unit.status = BlockedStatus("unable to install Opensearch Dashboards")
+            raise OSDInstallError(
+                "failed to install the Opensearch Dashboards snap. check logs for more details"
+            )
 
         # don't complete install until passwords set
         if not self.state.peer_relation:
--- a/src/charm.py
+++ b/src/charm.py
@@ -19,7 +19,6 @@ from events.oauth import OAuthHandler
 from events.requirer import RequirerEvents
 from events.tls import TLSEvents
 from events.upgrade import ODUpgradeEvents, OpensearchDashboardsDependencyModel
-from exceptions import OSDInstallError
 from helpers import (
     clear_global_status,
     clear_status,
@@ -132,11 +131,7 @@ class OpensearchDashboardsCharm(CharmBase):
         """Handler for the `on_install` event."""
         self.unit.status = MaintenanceStatus(MSG_INSTALLING)
 
-        if not self.workload.install():
-            self.unit.status = BlockedStatus("unable to install Opensearch Dashboards")
-            raise OSDInstallError(
-                "failed to install the Opensearch Dashboards snap. check logs for more details"
-            )
+        self.workload.install()
 
         # don't complete install until passwords set
         if not self.state.peer_relation:
--- a/src/charm.py
+++ b/src/charm.py
@@ -132,8 +132,7 @@ class OpensearchDashboardsCharm(CharmBase):
         """Handler for the `on_install` event."""
         self.unit.status = MaintenanceStatus(MSG_INSTALLING)
 
-        install = self.workload.install()
-        if not install:
+        if not self.workload.install():
             self.unit.status = BlockedStatus("unable to install Opensearch Dashboards")
             raise OSDInstallError(
                 "failed to install the Opensearch Dashboards snap. check logs for more details"
--- a/src/workload.py
+++ b/src/workload.py
@@ -11,7 +11,7 @@ import string
 import subprocess
 
 from charms.operator_libs_linux.v2 import snap
-from tenacity import retry
+from tenacity import retry, retry_if_result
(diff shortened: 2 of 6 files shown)


src/charm.py
```diff
@@ -134,6 +135,9 @@ def _on_install(self, event: InstallEvent) -> None:
         install = self.workload.install()
         if not install:
```
Mehdi-Bendriss:
```suggestion
        if not self.workload.install():
```

src/charm.py
```diff
@@ -134,6 +135,9 @@ def _on_install(self, event: InstallEvent) -> None:
         install = self.workload.install()
```
Mehdi-Bendriss: Can you also add a tenacity retry on the `install` method?

smethnani (dismissed):
Thanks @skourta!

src/workload.py
```diff
@@ -119,7 +119,12 @@ def healthy(self) -> bool:
...
 
     # --- Charm Specific ---
-
+    @retry(
+        stop=stop_after_attempt(3),
+        wait=wait_fixed(5),
+        reraise=True,
+        retry=retry_if_result(lambda result: result is False),
+    )
     def install(self) -> bool:
```
Mehdi-Bendriss: This shouldn't return a boolean, the return type should be None - we should rather catch the `SnapError` and raise an `OSDInstallError` or similar.

Mehdi-Bendriss (commented):
Thanks Smail, I left comment.

src/workload.py
gatici: WDYT on raising the custom errors when needed in the `def install` then catch, log and set status in the _on_install handler ?
   ```
       def install(self) -> None:
        """Loads the snap from LP, returning a StatusBase for the Charm to set.
        """

        cache = snap.SnapCache()
        try:
            dashboards = cache[self.SNAP_NAME]
        except KeyError as e:
            raise OSDInstallError(f"Snap {self.SNAP_NAME} not found in cache") from e
        try:
            dashboards.ensure(snap.SnapState.Present, revision=OPENSEARCH_DASHBOARDS_SNAP_REVISION)
            self.dashboards = dashboards
            self.dashboards.hold()


    def _on_install(self, event: InstallEvent) -> None:
        """Handler for the `on_install` event."""
        self.unit.status = MaintenanceStatus(MSG_INSTALLING)

        try:
            self.workload.install()
        except snap.SnapError as e:
            logger.error("OpenSearch Dashboards snap install failed: %s", str(e))
            self.unit.status = BlockedStatus("unable to install OpenSearch Dashboards")
            return
        except OSDInstallError as e:
            logger.error(str(e))
            self.unit.status = BlockedStatus("invalid snap state for installation")
            return
            .....
            
            
  ```

Mehdi-Bendriss: @gatici the problem with the above is that the juju lifecycle will continue on and assume there is an installed workload if we don't already crash the install hook. So it's better to prevent that by erroring the install hook

Mehdi-Bendriss (approved):
Thanks Smail!

smethnani (approved):

src/workload.py
gatici: OK.

gatici (approved):

Issue #177, opened 2025-04-22
Charm reaches blocked status when it fails to connect to OpenSearch


arturo-seijas:
Hi,

The current behavior of the charm is to [reach blocked status when it fails to connect to OpenSearch](https://github.com/canonical/opensearch-dashboards-operator/blob/2/edge/src/charm.py#L190).

It is my understanding that recovering from networking issues that might happen is something that the charm should do without requiring manual intervention from the operator so it should probably fall into waiting status instead.

Thanks for you time :)

PR #215, opened 2025-09-14
[DPE-7002] Update opensearch health checks

akram09:
## Description of issue or feature:
This PR addresses the issue mentioned in #177. After initial discussion, it was decided that the opensearch-dashboard charm should stay in a blocked status if the opensearch charm is not healthy or unreachable. This is due to the tight coupling between the two charms. The user is therefore asked to make interventions if the opensearch charm is not in a healthy state. At the current state, we don't make difference between opensearch being sel-healed or restarting and the case of opensearch not working correctly. 

Therefore this PR makes improvement over the opensearch health checks.

## Solution:
- Remove the `app_healthy` function and make calls directly to opensearch_ok  to avoid confusions.
- Add tenacity to the health checks. 
- Log the exception happening when json decode error occurs.
- Update charm status message when opensearch is in a yellow state.

## How was this change tested?
- [X] Manually
- [X] Unit tests
- [X] Integration tests

### DIFF
--- a/src/literals.py
+++ b/src/literals.py
@@ -51,7 +51,7 @@ MSG_STARTING_SERVER = "starting Opensearch Dashboards server..."
 MSG_WAITING_FOR_PEER = "waiting for peer relation"
 MSG_STATUS_DB_MISSING = "Opensearch connection is missing"
 MSG_STATUS_DB_DOWN = "Opensearch service is (partially or fully) down"
-MSG_STATUS_DB_UNHEALTHY = "Opensearch service is not in a green health state"
+MSG_STATUS_DB_UNHEALTHY = "The OpenSearch service health is red"
 MSG_TLS_CONFIG = "Waiting for TLS to be fully configured..."
 MSG_INCOMPATIBLE_UPGRADE = "Incompatible Opensearch and Dashboards versions"
 
@@ -65,6 +65,7 @@ MSG_STATUS_HANGING = "Application does not respond, request hanging"
 
 MSG_APP_STATUS = [
     MSG_STATUS_DB_DOWN,
+    MSG_STATUS_DB_UNHEALTHY,
 ]
 
 MSG_UNIT_STATUS = [
--- a/src/literals.py
+++ b/src/literals.py
@@ -51,6 +51,7 @@ MSG_STARTING_SERVER = "starting Opensearch Dashboards server..."
 MSG_WAITING_FOR_PEER = "waiting for peer relation"
 MSG_STATUS_DB_MISSING = "Opensearch connection is missing"
 MSG_STATUS_DB_DOWN = "Opensearch service is (partially or fully) down"
+MSG_STATUS_DB_UNHEALTHY = "Opensearch service is not in a green health state"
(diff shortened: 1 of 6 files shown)


src/literals.py
```diff
@@ -51,6 +51,7 @@
 MSG_WAITING_FOR_PEER = "waiting for peer relation"
 MSG_STATUS_DB_MISSING = "Opensearch connection is missing"
 MSG_STATUS_DB_DOWN = "Opensearch service is (partially or fully) down"
+MSG_STATUS_DB_UNHEALTHY = "Opensearch service is not in a green health state"
```
Mehdi-Bendriss: I would keep this one more generic, perhaps along the lines of:
```suggestion
MSG_STATUS_DB_UNHEALTHY = "The OpenSearch service health is red."
```

src/managers/health.py
```diff
@@ -78,42 +81,61 @@ def opensearch_ok(self) -> tuple[bool, str]:
...
+                f"\tException: {retry_state.outcome.exception()}"
+            )
+
         for endpoint in self.state.opensearch_server.endpoints:
+            retryer = Retrying(
+                stop=stop_after_attempt(3),
+                wait=wait_fixed(1),
+                reraise=True,
+                before_sleep=log_retry,
+            )
```
Mehdi-Bendriss: I think this would feel slightly more natural to directly set it in the retry context manager (as opposed to it being a variable)

src/managers/health.py
```diff
@@ -78,42 +81,61 @@ def opensearch_ok(self) -> tuple[bool, str]:
...
             if resp.status_code == 200:
                 try:
                     status = resp.json()
                 except requests.exceptions.JSONDecodeError:
+                    logger.error(f"Failed to decode JSON from {full_url}")
                     continue
+                if status.get("status") == "yellow":
+                    return False, MSG_STATUS_DB_UNHEALTHY
                 if status.get("status") == "green":
                     return True, ""
```
Mehdi-Bendriss: I don't think we should return False on Yellow, we should rather return False on Red. Perhaps:
```suggestion
                if status.get("status") == "red":
                    return False, MSG_STATUS_DB_UNHEALTHY
                    
                if status.get("status") in {"green", "yellow"}:
                    return True, ""
```

Mehdi-Bendriss (commented):
Thanks Akram, Looks good! I only have a couple of comments

smethnani (changes requested):
Thanks @akram09! I believe the db_unhealthy status message isn't being cleared when OpenSearch is green again, causing the `test_charm` and `test_process_interrupt` tests to fail. If you take a look at the [health check](https://github.com/canonical/opensearch-dashboards-operator/blob/773a644211fb457e1e77f0e7fb1f92393a41346f/src/charm.py#L213), perhaps the new message should also be cleared there or something to that effect.

smethnani (approved):
Great job @akram09! I tested this with opensearch and all messages are as expected and cleared when the health is not red or the opensearch service is available again.

Mehdi-Bendriss (approved):
Great work Akram!

akram09:
Thank you @Mehdi-Bendriss and @smethnani :raised_hands:

PR #217, opened 2025-09-18
[DPE-8018] add JWT authentication

reneradoi:
This PR adds JWT authentication to Opensearch Dashboards.

The changes have been tested manually, integration test coverage may be added once the [related PR](https://github.com/canonical/opensearch-operator/pull/696) in Opensearch has been merged and published.


src/literals.py
```diff
@@ -53,6 +54,7 @@
 MSG_STATUS_DB_DOWN = "Opensearch service is (partially or fully) down"
 MSG_TLS_CONFIG = "Waiting for TLS to be fully configured..."
 MSG_INCOMPATIBLE_UPGRADE = "Incompatible Opensearch and Dashboards versions"
+MSG_SECRETS_ACCESS_ERROR = "Failed to access secret, please check permissions"
```
Mehdi-Bendriss: this doesn't seem to be used

Mehdi-Bendriss (commented):
Thanks René! Minor comment on an used literal and integration tests. But the logic looks good

src/literals.py
reneradoi: Yeah true, I had added it before realizing we don't need to do secrets handling. Thanks for the hint.

Mehdi-Bendriss (approved):
Thank you René!

PR #224, opened 2025-12-04
Release 2.19.4 and Fixing Microk8s installation on SelfHosted Runners

akram09:
This Pull Request groups a set of separate PRs that aims to reach a stable release of OpenSearch, The included worked items are as follows:

### Terraform Bump to 1.0
This updates the juju terraform provider version to the latest version currently v1.0 . An important breaking change that comes with this version is the replacement of model with model_uuid in the juju_application resource. More informations can be found [here](https://github.com/juju/terraform-provider-juju/releases/tag/v1.0.0-rc1). The original [PR](https://github.com/canonical/opensearch-dashboards-operator/pull/221).

### Update OpenSearch Dashboards Artifacts
This updates the snap revisions and dependencies used by opensearch dashboards to `2.19.4` . The original [PR](https://github.com/canonical/opensearch-dashboards-operator/pull/222).

## Fixing spread jobs 
This fixes the for the issue we had in multiple charms regarding `chamcraft test --list github-ci` and replace it by installing spread as a golang package and use it to collect testing jobs. 

## Fixing OAuth Failing Tests 
This handles two main issues:
- First issue is fixing installation of microk8s on Self-Hosted runners. The tests were failing since microk8s wasn't able to start its containers due to dockerhub rate limiting. The solution would be to configure `docker.io/hosts.toml` to make it use the available `DOCKERHUB_MIRROR` that is already configured in self-hosted runners. 
- Second issue is regarding the integration with `identity-platform` bundle. The `test_oauth.py` was deploying the bundle from `latest/edge` which weirdly was using a new version of the `identity-platform-login-ui-operator` that is not working and tests are not passing with that version. The solution to that would be to add `iam-bundle.yaml` in our tests temporary till the identity team fixes this, and install the bundle from the yaml file.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -110,8 +110,9 @@ jobs:
           juju add-model test
       - name: Terraform deploy
         run: |
+          MODEL_UUID=$(juju show-model test | grep "model-uuid" | awk '{split($0, a, ": "); print a[2]}')
           pushd ./terraform
-          terraform apply -var='model=test' -auto-approve
+          terraform apply -var="model_uuid=$MODEL_UUID" -auto-approve
           popd
       - name: Wait for model and applications
         run: |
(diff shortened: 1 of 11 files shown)


skourta (approved):
Thank you @akram09 and @smethnani for the great work

PR #251, opened 2026-04-24
docs: Fix dashboards docs update workflow

izmalk:
Fixed the `docs/docs` directory path failure. We need to go from `docs/dashboards` upwards two levels, instead of just one. Replaced with `cd ../..`.
Fixed a potential issue with the CODEOWNERS file being empty. We check for its existence and only add the reviewers if the file exists.


izmalk:
Replace by https://github.com/canonical/opensearch-dashboards-operator/pull/252 to run the workflow manually

